PR #2, opened 2022-10-25
Variants

toolcreator:
Adds support for [CMake Variants](https://vector-of-bool.github.io/docs/vscode-cmake-tools/variants.html) from the [CMake Tools for Visual Studio Code](https://github.com/microsoft/vscode-cmake-tools).

It's far from perfect though currently (I'm a Lua beginner and didn't write any comments (yet)).
The YAML parsing is done by [simpleyaml](https://github.com/toolcreator/simpleyaml.lua) which I wrote specifically for this purpose.
At first I couldn't find any simple, dependency-free YAML parsing library. Then I found there are some options on github, but I was already done with my own implementation.
It does only parse well-formed files, probably only parses a subset of YAML, and doesn't have any error handling, so maybe it should be exchanged for something more robust.

A thing I added for this to work more fluently is, that CMakeSelectBuildType deletes the build directory if the chosen build type changes. This way it's quicker to change between build types.

variants.lua contains some CC-BY-SA licensed code from [rosettacode.org](https://rosettacode.org/wiki/Rosetta_Code), but that's [compatible with GPLv3](https://creativecommons.org/share-your-work/licensing-considerations/compatible-licenses/).

I'm dropping this here as an idea. If you're interested in integrating the feature we could discuss if we should change anything before merging.
Otherwise I may just keep it as is for my own use.

### DIFF
--- a/lua/simpleyaml.lua
+++ b/lua/simpleyaml.lua
@@ -24,15 +24,19 @@ SOFTWARE.
 
 local simpleyaml = {}
 
+-- parses the YAML file at `path` to a Lua table
+-- returns `nil` in case of error
 function simpleyaml.parse_file(path)
+  -- helper function to apply function `f(data)` at `nestingLevel` of `tab`
   local function atNestingLevel(nestingLevel, f, data, tab)
-    if nestingLevel == 0 then
+    if nestingLevel == 0 then -- arrived at `nestingLevel`, apply `f`
       f(data, tab)
-    else
+    else -- go deeper recursively
       atNestingLevel(nestingLevel - 1, f, data, tab[#tab].val)
     end
   end
 
+  -- helper function to insert a new `key` at `nestingLevel` of `tab`
   local function insertKey(key, nestingLevel, tab)
     atNestingLevel(
       nestingLevel,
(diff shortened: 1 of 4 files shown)


Civitasv:
Thanks, @toolcreator. It's really a nice feature. Things I think can be discussed:
1. As you said, you should add some comments.
2. About parsing YAML, can we just use `lua` format to store configurations of variants.

Civitasv:
> 2. About parsing YAML, can we just use `lua` format to store configurations of variants.
Or support both. And I just check your parser for yaml. I think it's ok, we can make it robust anyway.
What is your opinion?

lua/simpleyaml.lua
```diff
@@ -0,0 +1,121 @@
...
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+--]]
+
+local simpleyaml = {}
+
+function simpleyaml.parse_file(path)
```
Civitasv, line 27: Make it a member of simpleyaml: simpleyaml.parse_file = function(path)

toolcreator:
Thanks for the feedback!

## Notable changes

- Fix luacheck warnings
- Add comments to both extension code and simpleyaml
- Catch 1 more error condition in simpleyaml (more details below)

## On the input format

I think we should definitely support YAML, because the documentation for the variants makes heavy use of it.
The parsing is still a bit of an issue though: simpleyaml can parse simple and correct files, and from my perspective that's all you need for this feature. Yet, there is no nice feedback to the user in case of an erroneous input file. The extension would just do nothing, or apply a strange setting.
Making simpleyaml more robust would be a bigger task because it is line-based. To detect more errors I think a lot more state variables and/or a switch to character-based parsing would be required.
Unfortunately, it seems there are no(?) alternatives: Searching the web for YAML parsing sooner or later ends up with a YAML binding to some C library, and I wouldn't like adding more dependencies (especially complex non-Lua ones).
Searching for "lua yaml" on GitHub yields a few results that are rather old and also cannot parse the full feature set of YAML, so we might as well stick to my implementation.

The VSCode extension also supports JSON as an input format, and my implementation should do so as well via `vim.fn.json_decode`. I didn't bother testing this yet (at all) though.

## TODO

- [ ] Test JSON input
- [ ] Decide on YAML parsing
- [ ] ...?

Civitasv:
> To detect more errors I think a lot more state variables and/or a switch to character-based parsing would be required.

Yeah, I didn't find any lightweight YAML parsing library either. But I recently write a tiny JSON parser in C++. I think it's not that hard to swicth it to character-based parsing. I will try to make one when I have time.

Civitasv (approved):

Civitasv:
And I think we should also support lua format.

toolcreator:
> And I think we should also support lua format.

How would we go about with that? Read it from a file as well? Or do you have someting else in mind?

Civitasv:
Yes, I think it is okay to read it from a file, just like JSON and YAML format. It's natural to me.

PR #10, opened 2022-11-02
hot-fix:configration not work

lizhuoran1019:
Some configuration do not work. Eg. `cmake_console_size` ..
Because `utils.lua` and `config.lua` cannot properly load `const` variables that are changed by the `require("cmake-tools").setup()` in the neovim configuration


Civitasv:
Much appreciate for your pull request, but I prefer not change that much. The reason why some configuration not work is that I use multiple `const` variables. In fact, I should exactly have one instance of it. And I've fix this bug.
Please check the latest commit. Thanks anyway.

PR #13, opened 2022-11-03
Don't set env if you don't have it defined

lbiaggi:
I'm not sure what exactly happens, but plenary.job can't find the CMAKE_ROOT if env is nil 

My suggested change allows both cases to coexist.


Civitasv:
I think we have a better solution. I'll study how `Job:new` works.

PR #14, opened 2022-11-04
fix presets usage

lbiaggi:
New:
- `presets.lua`: I created a function that will parse CMakePresets.json and retrieve their name, since preset you need to pass the name, not the file
- `config.lua`: ah new nil parameter for preset value

Changes:
-`init.lua` - removed references for preset inside `clean`, since it doesn't really help with anything there,
-`init.lua` - changed how to select preset using `vim.ui.select` in `generate` and `build` functions, also I reordered the parameters since it's important to preset comes first than the others.

### DIFF
--- a/lua/cmake-tools/config.lua
+++ b/lua/cmake-tools/config.lua
@@ -14,7 +14,8 @@ local Config = {
   build_type = nil,
   build_target = nil,
   launch_target = nil,
-  cmake_kit = nil
+  cmake_kit = nil,
+  preset = nil,
 }
 
 function Config:new(const)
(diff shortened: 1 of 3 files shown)


Civitasv:
Thanks for your pull request, I think it should work as expected now.

Civitasv:
Maybe, We should split configure preset and build preset?

lbiaggi:
I agree, I forgot about the possibility they aren't 1:1, maybe as additional flag, “do you want to reuse <preset> used to configure as build Preset?” something like that

Civitasv:
I split them. See the latest commit.

lbiaggi:
Yeah, but I think, it's worth to mention the possibility of define a default one?
e.g: 
```
   cmake_build_preset = "prod"
   cmake_configure_preset = "dev"
  }
```

Civitasv:
I don't think so. This option may confuse people who don't use `CMakePresets`.

Issue #30, opened 2023-02-10
Handle Preset include


CarlosNihelton:
**Describe the problem or limitation you are having**

`CMakePresets.json` and `CMakeUserPresets.json` can include other files with the include field in file version 4 and later. Files included by these files can also include other files. If `CMakePresets.json` and `CMakeUserPresets.json` are both present, `CMakeUserPresets.json` implicitly includes `CMakePresets.json`, even with no include field, in all versions of the format.
Users may leverage that feature to have common definitions across different projects.

See [cmake-preset docs](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#includes)

The conan package manager uses that when generating multiple build configurations, for example:

`CMakeUserPresets.json`:

```json
{
    "version": 4,
    "vendor": {
        "conan": {}
    },
    "include": [
        "~/project/build/CMakePresets.json"
    ]
}
```

`~/project/build/CMakePresets.json`:

```json
{
    "version": 3,
    "vendor": {
        "conan": {}
    },
    "cmakeMinimumRequired": {
        "major": 3,
        "minor": 15,
        "patch": 0
    },
    "configurePresets": [
        {
            "name": "default",
            "displayName": "'default' config",
            "description": "'default' configure using 'Ninja Multi-Config' generator",
            "generator": "Ninja Multi-Config",
            "cacheVariables": {
                "CMAKE_POLICY_DEFAULT_CMP0091": "NEW"
            },
            "toolchainFile": "~/project/build/conan_toolchain.cmake",
            "binaryDir": "~/project/build"
        }
    ],
    "buildPresets": [
        {
            "name": "release",
            "configurePreset": "default",
            "configuration": "Release"
        },
        {
            "name": "debug",
            "configurePreset": "default",
            "configuration": "Debug"
        },
        {
            "name": "relwithdebinfo",
            "configurePreset": "default",
            "configuration": "RelWithDebInfo"
        }
    ],
    "testPresets": [
        {
            "name": "release",
            "configurePreset": "default",
            "configuration": "Release"
        },
        {
            "name": "debug",
            "configurePreset": "default",
            "configuration": "Debug"
        },
        {
            "name": "relwithdebinfo",
            "configurePreset": "default",
            "configuration": "RelWithDebInfo"
        }
    ]
}

```
The plugin current behavior cannot detect the presets coming from the included file.

**Describe the solution you'd like**

[...]

PR #31, opened 2023-02-10
Handle includes in the top level presets file

CarlosNihelton:
Here is a suggestion to partially tackle #30 .

The proposed changes follow the files included in the top level `CMake[User]Presets.json` file, decodes them and merge their presets into the table that will be later used when the plugin checks for presets.

Thank you for the awesome plugin!


Civitasv:
Thanks for your implementation for this feature. Very useful. Can you change `mergeTableListByKey` to `merge_table_list_by_key` so we can make code more consistent?

Issue #32, opened 2023-02-20
Cannot run an executable, if its path contains spaces


eaxecx:
**Bug description**
I have a project on my external hard drive, and the drive is mounted as `/run/media/eax_ebx/Verbatim Drive`. I can successfully generate and build my project, but when I run `:CMakeRun` or click the Run button on lualine, select the executable, it gives the error `zsh:1: no such file or directory: /run/media/eax_ebx/Verbatim`. This error does not appear, if the path to the executable does not contain spaces.

**Steps to reproduce**

1. Create a directory like `/home/<username>/my project`.
2. Add a CMakeLists.txt:
```cmake
cmake_minimum_required(VERSION 3.15)

project(test LANGUAGES CXX)

add_executable(test "Test.cpp")
```

3. Add the Test.cpp:
```cpp
#include <iostream>
#include <cstdlib>

int main() {
    std::cout << "Hello, World!\n";

    return EXIT_SUCCESS;
}
```

4. Configure, build the `test` project and run it. On my machine, I get the error `zsh:1: no such file or directory: /home/eax_ebx/my`.
5. Now rename the project directory from `my project` to `my_project` and delete the build directory. Repeat the 4. step. But now, on my machine, I don't get the error and it successfully runs and I see "Hello, World!" in the console.

**Expected behavior**
I think no matter what the path to the executable looks like (contains Unicode symbols, whitespaces, etc.), the plugin should run this executable without any problems.

**Environment**

- OS: ArchLinux, kernel version is 6.1.12-arch1-1. My shell is zsh 5.9 (x86_64-pc-linux-gnu).
- Desktop environment: Window Maker 0.95.9
- Plugins commit hash: cda96e7a64590c23bad28b3a96d195726f70ed4b

**Additional context**
Configuration of cmake-tools in init.vim:
```vim
Plug 'Civitasv/cmake-tools.nvim'

lua << EOF
require("cmake-tools").setup {
  cmake_command = "cmake",
  cmake_build_directory = "build",
  cmake_build_directory_prefix = "", -- when cmake_build_directory is "", this option will be activated
  cmake_generate_options = { "-D", "CMAKE_EXPORT_COMPILE_COMMANDS=1", },
  cmake_build_options = {},
  cmake_console_size = 20, -- cmake output window height
  cmake_console_position = "belowright", -- "belowright", "aboveleft", ...
  cmake_show_console = "always", -- "always", "only_on_error"
[...]

PR #33, opened 2023-02-20
Update init.lua

eaxecx:
Wrapped the variable target_path in quotes in cmake.run(), so that the executable could run even if its path contains spaces.


eaxecx:
#32

Civitasv:
Thanks.

PR #38, opened 2023-04-12
Added Naming Convention

rohit-kumar-j:
Tested with both `cmake-kits.json` and `CMakeKits.json`

### DIFF
--- a/lua/cmake-tools/kits.lua
+++ b/lua/cmake-tools/kits.lua
@@ -8,7 +8,7 @@ function kits.parse()
     local files = vim.fn.readdir(".")
     local file = nil
     for _, f in ipairs(files) do -- iterate over files in current directory
-      if f == "cmake-kits.json" then -- if a kits config file is found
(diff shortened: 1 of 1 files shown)


Civitasv:
Why is it useful?

Civitasv:
VsCode's cmake-tools only supports `cmake-kits.json`. If our plugin also supports `CMakeKits.json`, then users can't use both neovim and vscode. I think it is not suitable. What do you think about it?

rohit-kumar-j:
Oh, I didn't know about that!😅. I thought that VS Code supported `CMakeKits.json` since the convetions were caplitalized letters in `CMakeLists.txt` and `CMakeUserPresets.json`,etc and the readme had it caplitalized.

> Attention: If CMake[User]Presets.json is provided, then CMakeKits.json or CMakeVariants.[json|yaml] won't have any effect.

However, wouldn't it be alright since this supports both?:
```lua
 if (f == "cmake-kits.json" or f == "CMakeKits.json")  
```

Civitasv:
You've convinced me.

PR #43, opened 2023-04-27
Add hooks for clangd and ccls to get compile command directory from preset

Chemiseblanc:
This adds an alternative to soft-linking compile_comands.json by adding functions that users can call in the on_new_config callback for clangd or ccls when using lspconfig.
It sets the compilation database directory to their current preset build directory and updates this path on regen.

The only existing behavior that had to be modified is in utils.show_cmake_console() where it now returns focus to the previous window on open. This is done because when restarting the language server it incorrectly uses the current buffer as the attach target instead of the previously attached buffer.

### DIFF
--- a/lua/cmake-tools/init.lua
+++ b/lua/cmake-tools/init.lua
@@ -639,8 +639,8 @@ function cmake.has_cmake_preset()
 end
 
 function cmake.configure_compile_commands()
-  if config.lsp_type == nil then
-    if config.cmake_soft_link_compile_commands then
+  if const.lsp_type == nil then
+    if const.cmake_soft_link_compile_commands then
       cmake.compile_commands_from_soft_link()
     end
   else
@@ -649,7 +649,7 @@ function cmake.configure_compile_commands()
 end
 
 function cmake.compile_commands_from_soft_link()
-  if config.build_directory == nil or config.lsp_type ~= nil then return end
+  if config.build_directory == nil or const.lsp_type ~= nil then return end
 
   local source = config.build_directory.filename .. "/compile_commands.json"
   local destination = vim.loop.cwd() .. "/compile_commands.json"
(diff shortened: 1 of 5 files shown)


Civitasv:
Fantastic! Can you also add a section in README on how to use this feature?

Chemiseblanc:
I added a section with an example for clangd. Ccls has the same usage with the names changed.

lua/cmake-tools/init.lua
```diff
@@ -133,13 +128,6 @@ function cmake.generate(opt, callback)
...
         callback()
       end
-
-      if const.cmake_soft_link_compile_commands then
-        utils.softlink(
-          config.build_directory.filename .. "/compile_commands.json",
-          vim.loop.cwd() .. "/compile_commands.json"
-        )
-      end
     end,
```
Civitasv, line 131: Why don't we have `cmake.configure_compile_commands()` here?

Chemiseblanc: You're right. I've added it.

rohit-kumar-j:
@Chemiseblanc, is there a way to add this within the `init.lua`? Something like this:

```lua
vim.api.nvim_set_var('lsp_servers',
(
    {
        name = 'clangd',
        cmd = {
            "clangd",
        },
        root_dir = function(fname)
            return lspconfig.util.find_git_ancestor(fname)
        end,
        on_new_config = [[settings here...]] 
        settings = {
            clangd = {
                extraArgs = {}
            },
        },
    },
    {
        name = 'pyright',
    },
  -- other servers and settings
)
```

Reasoning: I have a table to servers that are seperate from the function calls, i.e., no seperate `require.lspserver().clangd` calls for each of the filetypes/projects.

```lua
for _, lsp in ipairs(lsp_servers) do
    require('lspconfig')[lsp.name].setup {
        settings = lsp_servers[lsp.settings],
        on_attach = on_attach,
        capabilities = capabilities,
        handlers = handlers,
    }
end
```

Would this be possible to implement/can this be done on attach instead of on new config?

Civitasv:
This is really useful.

Chemiseblanc:
The reason for the callbacks was because lspconfig will overwrite the existing settings if the setup function is called more than once and I couldn't find a different way to modify the language server settings.

If this works without interfering with the existing config, I agree it's a better way to support this feature.

rohit-kumar-j:
@Chemiseblanc, this may not be relevant to cmaketools, do you know how to configure the `.clang-format.txt` in root directory with lsp?

Civitasv:
Maybe "--fallback-style=<string>           - clang-format style to apply by default when no .clang-format file is found" is helpful?

rohit-kumar-j:
do you have an example confg?

PR #47, opened 2023-05-11
CleanReBuild, Remove generate on build

rohit-kumar-j:
1. Added Clean Rebuild, which cleans project and builds target
2. Remove 'hack': {generate each time on build}. Ref: [for dependencies handling build targets](https://discourse.cmake.org/t/ability-to-clean-individual-targets/2980). Must generate only once and not each time unless the CMakeLists.txt is modified. Why?: Configure/Generate each time is takes too long for large projects(even if it only checks save timestamps for detecting file modifications).

This is still experimental, so experimental is set to 0 in lua/cmake-tools/init.lua

- [x] Tested on Windows 10 MSVC, [MSBUILD, Ninja]
- [x] Tested on Linux/WSL, [Unix MakeFiles]
- [ ] Tested on MacOS

Testing No.2 with autocmd which saves only if buffer is modified:
(Currently this does not work, as it always throws buf_modified = false, perhaps a fault in logic)
Hoping somone can help me with this:

<details>
<Summary>Autocmd to save on modified CMakeLists.txt</Summary>

commit : [1f0404e](https://github.com/Civitasv/cmake-tools.nvim/pull/47/commits/1f0404ee664a7c56c0ae1224955ecdf935c6b39a)

```lua

-- init.lua : 

-- preload the autocmd if the following option is true. only saves cmakelists.txt files
if const.cmake_regenerate_on_save == true then
  vim.api.nvim_create_autocmd("BufWritePost", {
    group = vim.api.nvim_create_augroup("cmaketools", {clear = true}),
    pattern  = "CMakeLists.txt",
    callback = function()
      -- local fargs = opt.fargs or {}
      local buf = vim.api.nvim_get_current_buf()
      -- local buf_modified  = vim.api.nvim_get_all_options_info()['modified']
      local buf_modified  = vim.api.nvim_get_all_options_info()['modified']
      -- print("buf_modified: " .. utils.dump(buf_modified))
      print("buf: " .. buf .. ", mod: " .. utils.dump(buf_modified))
      if not vim.api.nvim_buf_get_option(buf, "modified") then
        return
      end
      -- vim.api.nvim_buf_call(buf, function()
      --   cmake.generate({ bang = false, fargs = {} },
      --     function()
      --       -- no function here
      --     end)
      -- end)
        vim.schedule(
          function()
            cmake.generate({ bang = false, fargs = {} },
              function()
                -- no function here
              end)
          end)
        -- print("buffer is not modified... not saving!")
    end,
  })
end
```
</details>


Civitasv:
Thanks, I will test it on Linux later.

Civitasv:
> Testing No.2 with autocmd which saves only if buffer is modified:
> (Currently this does not work, as it always throws buf_modified = false, perhaps a fault in logic)
> Hoping someone can help me with this:

It should be `BufWritePre` instead of `BufWritePost` cause `BufWritePost` cmd will only be activated after saving entire buffer to file, so of course buffer is not modified.

rohit-kumar-j:
@Civitasv, This(suffix version) runs the CMakeRun command in the buffer but does not keep the `CMakeRun` buffer alive for scrolling the output after execution:

```lua
  
  -- utils.lua
  
  if const.cmake_launch_from_build_directory == true then
    local suffix = string.format(" cd %s && %s", opts.cmake_launch_path, executable)
    vim.cmd(prefix .. " | term " .. suffix)
  else
    -- vim.cmd(prefix .. " | term " .. executable)
  end
```

unlike:

```lua
vim.cmd(prefix .. " | term " .. executable)
```

I cannot seem to find the error...

lua/cmake-tools/config.lua
```diff
@@ -1,10 +1,12 @@
 local Path = require("plenary.path")
 local scandir = require("plenary.scandir")
 local Result = require("cmake-tools.result")
-local utils = require("cmake-tools.utils")
+-- local utils = require("cmake-tools.utils") -- Fails lua check. Uncomment this for testing
 local Types = require("cmake-tools.types")
 local variants = require("cmake-tools.variants")
 
+local vim = vim -- Localize LSP errors to a single line: [Undefined global 'vim']
```
Civitasv, line 8: maybe we don't need this, you can change your `sumneko` settings to:
```lua
return {
  settings = {
    ...
    Lua = {
      diagnostics = {
        globals = { "vim" },
        neededFileStatus = {
          ["codestyle-check"] = "Any",
        },
      },
    },
  },
}
```

rohit-kumar-j: Ah yes! the global vim.

rohit-kumar-j: This error is still persistant. I think I am using lua_ls, not sumeko is there another option to disable this?

Civitasv: I don't know how to configure with lua_ls, maybe there is an option for it.

lua/cmake-tools/init.lua
```diff
@@ -218,10 +216,90 @@ function cmake.build(opt, callback)
...
+--- Clean Rebuild: Clean the project and then Rebuild the target
+--- [See dependancy discussion here]
+function cmake.clean_rebuild(opt, callback)
+  if not utils.has_active_job() then
+    return
+  end
+
+  -- Check of project is configured
+  if config.build_directory == nil then
+    local fargs = fargs or opt.fargs
```
Civitasv, line 276: maybe it should be: `local fargs = opt.fargs or {}`?

lua/cmake-tools/init.lua
```diff
@@ -273,6 +351,11 @@ function cmake.open()
   utils.show_cmake_console(const.cmake_console_position, const.cmake_console_size)
 end
 
+local getPath = function(str,sep)
+    sep=sep or'/'
```
Civitasv, line 355: format stuff:
sep = sep or '/'

rohit-kumar-j:
> It should be BufWritePre instead of BufWritePost cause BufWritePost cmd will only be activated after saving entire buffer to file, so of course buffer is not modified. 

I see. That is the result if it is modified. The can we do something like this:

```lua
if not buf_modified then
    -- save run generate command
end
```

lua/cmake-tools/init.lua
```diff
@@ -304,19 +387,25 @@ function cmake.run(opt, callback)
       vim.schedule(function()
         result = config:get_launch_target()
         -- print(utils.dump(result))
-        local target_path = result.data
         -- print("TARGET", target_path)
+        local target_path = result.data
         local is_win32 = vim.fn.has("win32")
         if (is_win32 == 1) then
           -- Prints the output in the same cmake window as in wsl/linux
+          local new_s= getPath(target_path, '/')
```
Civitasv, line 395: format.

lua/cmake-tools/init.lua
```diff
@@ -697,4 +786,32 @@ function cmake.ccls_on_new_config(new_config)
   new_config.init_options.compilationDatabaseDirectory = config.build_directory.filename
 end
 
+-- preload the autocmd if the following option is true. only saves cmakelists.txt files
+if const.cmake_regenerate_on_save == true then
+  vim.api.nvim_create_autocmd("BufWritePost", {
```
Civitasv, line 791: BufWritePre

lua/cmake-tools/init.lua
rohit-kumar-j: done

lua/cmake-tools/init.lua
rohit-kumar-j: done

lua/cmake-tools/init.lua
rohit-kumar-j: done

Civitasv:
> but does not keep the `CMakeRun` buffer alive for scrolling the output after execution:

You mean when you press `enter`, it disappear?

lua/cmake-tools/init.lua
rohit-kumar-j: done

rohit-kumar-j:
> > but does not keep the `CMakeRun` buffer alive for scrolling the output after execution:
> 
> You mean when you press `enter`, it disappear?


yes. or anykey. not just enter.

Civitasv:
It used to be that way too. I mapped `A-x` to change terminal mode to normal mode:
`keymap("t", "<A-x>", [[<C-\><C-n>]], { silent = true })`

rohit-kumar-j:
>>>> It should be BufWritePre instead of BufWritePost cause BufWritePost cmd will only be activated after saving entire buffer to file, so of course buffer is not modified.
>>>
>>> I see. That is the result if it is modified. The can we do something like this:
>>
>>if not buf_modified then
>>    -- save run generate command
>>end
>>
>yes

Still same behaviour

Civitasv:
> Still same behaviour.

Weird, it works for me.

rohit-kumar-j:
> keymap("t", "<A-x>", [[<C-\><C-n>]], { silent = true })

Is this `Alt + x` ?

Civitasv:
> Is this `Alt + x` ?

Yes.

rohit-kumar-j:
What does it do?

Civitasv:
> What does it do?

It changes current mode from terminal mode to normal mode.

rohit-kumar-j:
I see now. But how does this help with the buffer not persisting an output?

rohit-kumar-j:
> > Still same behaviour.
> 
> Weird, it works for me.

My inention was to CMakeGenerate each time on save: if file_type==CMakeLists.txt and if file_is_modified.  


Is this same snippet?

```lua
if const.cmake_regenerate_on_save == true then
  vim.api.nvim_create_autocmd("BufWritePre", {
    group = vim.api.nvim_create_augroup("cmaketools", {clear = true}),
    pattern  = "CMakeLists.txt",
    callback = function()
      local buf = vim.api.nvim_get_current_buf()
      -- TODO: do some logic here to check if buffer is actually modified, and only if it is modified,
      -- execute the generate, otherwise return. This is to avoid unnecessary regenerattion
      local buf_modified  = vim.api.nvim_buf_get_option(buf, 'modified')
      if not buf_modified then
        vim.schedule(
          function()
            cmake.generate({ bang = false, fargs = {} },
              function()
                -- no function here
              end)
          end)
        end
      -- print("buffer is not modified... not saving!")
    end,
  })
end

```

Civitasv:
> ```lua
> if not buf_modified then
> ```

It should be `if buf_modified`

Civitasv:
> I see now. But how does this help with the buffer not persisting an output?

In normal mode, it won't disappear when you touch `enter` or other key. unless you use `:q`

rohit-kumar-j:
> It should be if buf_modified

Even with this, if I modify the buffer and save it does not CMakeGenerate.

rohit-kumar-j:
OMG! this is embarassing, I had the `cmake_regenerate_on_save` in const set to false. It works as expected now.

Civitasv:
> OMG! this is embarassing, I had the `cmake_regenerate_on_save` in const set to false. It works as expected now.

Okay. :smile:

lua/cmake-tools/variants.lua
Civitasv, line 1: Why do you change code format of this file?

rohit-kumar-j:
There is one final piece:

```lua
-- utils.lua

function utils.execute(executable, opts)
  -- save all
  vim.cmd("wall")
  -- print("EXECUTABLE", executable)
  local set_bufname = "file '" .. opts.bufname .. "'"
  print(set_bufname)

  local prefix = string.format("%s %d new", opts.cmake_console_position, opts.cmake_console_size)
  utils.close_cmake_console();

  -- This launches the executable target from its exact build directory. Useful if executable produces an out.txt file.
  if const.cmake_launch_from_built_binary_directory == true then
    local suffix = string.format(" cd %s && %s", opts.cmake_launch_path, executable)
    -- local suffix = " cd " .. opts.cmake_launch_path .. " && " .. executable
    vim.cmd(prefix .. " | term " .. suffix)
  else
    vim.cmd(prefix .. " | term " .. executable)
  end

  vim.opt_local.relativenumber = false
  vim.opt_local.number = false
  vim.cmd(set_bufname) -- preserves behaviour
  vim.bo.buflisted = false
  vim.cmd("startinsert")

end

```

The suffix method does not keep the output form the executable upon launch from binary directory. Would there be a suitable fix for this or do you think we should approach this differently?

lua/cmake-tools/variants.lua
rohit-kumar-j: of which file?

Civitasv:
> The suffix method does not keep the output form the executable upon launch from binary directory.

I don't quite understand. Can you give a specific example?

rohit-kumar-j:
Give me second. I will record and send a screen capture.

rohit-kumar-j:
It does not let me move the mouse output after `process exited with code 0`:

I pressed the left arrow key and it erased the buffer.

https://github.com/Civitasv/cmake-tools.nvim/assets/37873142/33fc36c3-5973-4027-9edc-dbe8da6eb42c

rohit-kumar-j:
If I have the deafults on, I get this error:

image

And then it lets me scroll the process output buffer.

Civitasv:
Why can I?
Peek 2023-05-12 11-10

Civitasv:
> I get this error:

That's because neovim cannot have two buffers which have same name.

rohit-kumar-j:
> If I have the deafults on, I get this error:
> 
> image
> 
> And then it lets me scroll the process output buffer. Like this:
> 
>  cmake_tools_nvim_scrolling.mp4

This is the config: `cmake_launch_from_built_binary_directory = false,`

rohit-kumar-j:
> See video at very end:
> 
> It does not let me move the mouse output after `process exited with code 0`:
> 
> I pressed the left arrow key and it erased the buffer.
> 
>  problem_cmake_tools.mp4

This is the config: `cmake_launch_from_built_binary_directory = true,`

rohit-kumar-j:
Is it because of `vim.cmd(set_bufname)`?

Civitasv:
> Is it because of `vim.cmd(set_bufname)`?

Yes.

Civitasv:
> This is the config: `cmake_launch_from_built_binary_directory = true,`

I also set it to true.

rohit-kumar-j:
> > I get this error:
> 
> That's because neovim cannot have two buffers which have same name.

Can we do `a vim.api.nvim_cmd(cmd, opts)` instead of `vim.cmd()` this, or keep a persistent output buffer through out the life of the CMakeTools lifetime that continuously writes to an output buffer like in VSCode. Acts sort of like a log file?

Civitasv:
> Can we do `a vim.api.nvim_cmd(cmd, opts)` instead of `vim.cmd()` this, or keep a persistent output buffer through out the life of the CMakeTools lifetime that continuously writes to an output buffer like in VSCode. Acts sort of like a log file?

That's a way.

rohit-kumar-j:
Should I undo the changes in utils.lua for now so that we can push the changes so far?

lua/cmake-tools/variants.lua
Civitasv: variants.lua

Civitasv:
> Should I undo the changes in utils.lua for now so that we can push the changes so far?

Okay.

rohit-kumar-j:
Also, should I remove the experimental from `cmake.build()` to make it permanent? Can you test this?

Civitasv:
> Also, should I remove the experimental from `cmake.build()` to make it permanent? Can you test this?
> 
> (Since it autogenerates with `BufWritePre` when CMakeLists.txt is saved?)

I’ve tested it. It works as expected.

rohit-kumar-j:
Okay I will refactor it and push.

rohit-kumar-j:
Ah! I see

rohit-kumar-j:
Done.

rohit-kumar-j (commented):
Need to solve for persisting the data in the output buffer upon execution of executable targets.

rohit-kumar-j:
@Civitasv, [neodev](https://github.com/folke/neodev.nvim) solves the `Undefined global 'vim'` issue, plus as a bonus gives signature support. 

### Steps for future readers:
Add an empty `lua/lua/.luarc.json` in your nvim config and run setup before lspconfig: `require('noedev').setup()`

Preview:

image

image

Issue #34, opened 2023-02-27
Set working directory for running


TheSunCat:
**Describe the problem or limitation you are having**
I am working on a C++ project which accesses files in a `Resources` folder as if they were in the working directory. When running the project through `CMakeRun`, the working directory is that of nvim, so it cannot load the files it needs.

**Describe the solution you'd like**
The ability to configure the working directory used when launching the program.

**Describe alternatives you've considered**
I found [this](https://github.com/microsoft/vscode-cmake-tools/issues/1395#issuecomment-694968338) issue in vscode-cmake-tools, but setting `cwd` in the `require("cmake-tools").setup` call did not appear to do anything. I might be doing this wrong, however, as I'm not sure how `launch.json` settings translate to this project.

Civitasv:
Can you give a specific example? like the file structure of your project

TheSunCat:
Yes of course. I am using [this](https://github.com/samynk/GameDevEngine2) project by adding `DefineGameProject(Test)` at the end of CMakeLists.txt, running cmake once to generate it, then running `nvim Projects/Test/Test.cpp`. The workring directory would ideally be `$(pwd)/Projects/$project/Resources`, so each project can load its own resources, but it appears to be set to the path nvim was invoked from.

rohit-kumar-j:
@Civitasv, I faced a similar limitation. My executable was in `build/bin/Debug/exec.exe`, and it is supposed to produce a `data.txt` file.

When `CMakeRun` is invoked, it produces a file in the directory from which it was run and produces outputs in the same directory (`root_dir/data.txt`) which is undesirable.(`root_dir ~$ build/bin/Deubg/exec.exe`)

On VSCode, when `CMakeRun` is invoked, the executable is run from its own build directory by default: (`root_dir/build/bin/Deubg ~$ exec.exe`) thereby not cluttering the root_dir. The output files are in (`root_dir/build/bin/Debug/data.txt`)

I think a conditional default option to set this behaviour would be a good idea:..

```lua

-- const.lua

const = {
    ...
    cmake_launch_from_build_dir = true,
    ...
}
```

rohit-kumar-j:
[...]

PR #51, opened 2023-05-15
Launch from build directory fix, and can scroll output

rohit-kumar-j:
Solves https://github.com/Civitasv/cmake-tools.nvim/pull/47#issuecomment-1545068911

The intended behaviour for #34(on by default), but also keeps the old behaviour :D

### DIFF
--- a/lua/cmake-tools/utils.lua
+++ b/lua/cmake-tools/utils.lua
@@ -68,41 +68,27 @@ function utils.execute(executable, opts)
   -- save all
   vim.cmd("wall")
   -- print("EXECUTABLE", executable)
-  if const.cmake_launch_from_built_binary_directory then
-    utils.job = Job:new({
-      command = executable,
-      args = {},
-      cwd = opts.cmake_launch_path,
-      on_stdout = vim.schedule_wrap(append_to_cmake_console),
-      on_stderr = vim.schedule_wrap(append_to_cmake_console),
-      on_exit = vim.schedule_wrap(function(_, code, signal)
-        append_to_cmake_console("Exited with code " .. (signal == 0 and code or 128 + signal))
-        if code == 0 and signal == 0 then
-          if opts.on_success then
-            opts.on_success()
-          end
-        elseif opts.cmake_show_console == "only_on_error" then
-          utils.show_cmake_console(opts.cmake_console_position, opts.cmake_console_size)
-          vim.api.nvim_command("cbottom")
+  utils.job = Job:new({
+    command = executable,
+    args = {},
+    cwd = opts.cmake_launch_path,
+    on_stdout = vim.schedule_wrap(append_to_cmake_console),
+    on_stderr = vim.schedule_wrap(append_to_cmake_console),
+    on_exit = vim.schedule_wrap(function(_, code, signal)
+      append_to_cmake_console("Exited with code " .. (signal == 0 and code or 128 + signal))
+      if code == 0 and signal == 0 then
+        if opts.on_success then
+          opts.on_success()
         end
-      end),
-    })
+      elseif opts.cmake_show_console == "only_on_error" then
+        utils.show_cmake_console(opts.cmake_console_position, opts.cmake_console_size)
+        vim.api.nvim_command("cbottom")
+      end
+    end),
+  })
 
-    utils.job:start()
-    return utils.job
-  else
-    local set_bufname = "file " .. opts.bufname
-    local prefix = string.format("%s %d new", opts.cmake_console_position, opts.cmake_console_size)
-    utils.close_cmake_console();
-    vim.cmd(prefix .. " | term " .. executable)
-    vim.opt_local.relativenumber = false
-    vim.opt_local.number = false
-    vim.cmd(set_bufname)
-    vim.bo.buflisted = false
-    vim.cmd("startinsert")
-  end
-
-
+  utils.job:start()
+  return utils.job
 end
 
 function utils.softlink(src, target)
--- a/README.md
+++ b/README.md
@@ -102,7 +102,6 @@ require("cmake-tools").setup {
   cmake_build_directory_prefix = "cmake_build_", -- when cmake_build_directory is "", this option will be activated
   cmake_generate_options = { "-D", "CMAKE_EXPORT_COMPILE_COMMANDS=1" },
   cmake_regenerate_on_save = true, -- Saves CMakeLists.txt file only if mofified.
-  cmake_launch_from_built_binary_directory = true, -- WIP: see #47 and #34
   cmake_soft_link_compile_commands = true, -- if softlink compile commands json file
   cmake_build_options = {},
   cmake_console_size = 10, -- cmake output window height
--- a/lua/cmake-tools/const.lua
+++ b/lua/cmake-tools/const.lua
@@ -4,7 +4,6 @@ local const = {
   cmake_command = "cmake",                                          -- cmake command path
   cmake_generate_options = { "-DCMAKE_EXPORT_COMPILE_COMMANDS=1" }, -- it will be activated when invoke `cmake.generate`
   cmake_regenerate_on_save = true,
-  cmake_launch_from_built_binary_directory = true,
   cmake_soft_link_compile_commands = true,
   cmake_compile_commands_from_preset = false,
   cmake_build_options = {},              -- it will be activated when invoke `cmake.build`


lua/cmake-tools/utils.lua
```diff
@@ -68,16 +68,41 @@ function utils.execute(executable, opts)
...
-  local prefix = string.format("%s %d new", opts.cmake_console_position, opts.cmake_console_size)
-
-  utils.close_cmake_console();
-  vim.cmd(prefix .. " | term " .. executable)
-  vim.opt_local.relativenumber = false
-  vim.opt_local.number = false
-  vim.cmd(set_bufname)
-  vim.bo.buflisted = false
-  vim.cmd("startinsert")
+  if const.cmake_launch_from_built_binary_directory == true then
```
Civitasv, line 71: No need to check `== true`, just use `if const.cmake_launch_from_built_binary_directory`

lua/cmake-tools/utils.lua
```diff
@@ -68,16 +68,41 @@ function utils.execute(executable, opts)
...
+          utils.show_cmake_console(opts.cmake_console_position, opts.cmake_console_size)
+          vim.api.nvim_command("cbottom")
+        end
+      end),
+    })
+
+    utils.job:start()
+    return utils.job
+  else
+    local set_bufname = "file " .. opts.bufname
```
Civitasv, line 94: I think maybe we can delete this chunk, and use cmake_launch_from_built_binary_directory by default.

rohit-kumar-j: fully remove the `vim.cmd()` part?

rohit-kumar-j: If so, then how about, remove this from the `const` table itself and always launch from built binary directory?

Civitasv: Yes, that's what I mean.

rohit-kumar-j: Done.

rohit-kumar-j:
Modified Readme to suit this.

Civitasv:
Forget to say, have you test if we can input for `cin` in the terminal when run the executable?

rohit-kumar-j:
> Forget to say, have you test if we can input for `cin` in the terminal when run the executable?

input for cin? When the program is running?

Civitasv:
Yes. For example:

```cpp
int main()
{
    int a;
    cin << a;
    cout << "A: " << a << endl;
    return 0;
}
```

When I run this program, I should be able to input.

rohit-kumar-j:
Yes, there seems to some [documentation](https://neovim.io/doc/user/job_control.html) on this, although, not in lua. Can you suggest a change for `cin`? 

This does not seem to work:
```lua
function utils.execute(executable, opts)
  -- save all
  vim.cmd("wall")
  -- print("EXECUTABLE", executable)
  utils.job = Job:new({
    command = executable,
    args = {},
    cwd = opts.cmake_launch_path,
    stdin = "pipe",
    on_stdout = vim.schedule_wrap(append_to_cmake_console),
    on_stderr = vim.schedule_wrap(append_to_cmake_console),
    on_exit = vim.schedule_wrap(function(_, code, signal)
      append_to_cmake_console("Exited with code " .. (signal == 0 and code or 128 + signal))
      if code == 0 and signal == 0 then
        if opts.on_success then
          opts.on_success()
        end
      elseif opts.cmake_show_console == "only_on_error" then
        utils.show_cmake_console(opts.cmake_console_position, opts.cmake_console_size)
        vim.api.nvim_command("cbottom")
      end
    end),
  })

```

Civitasv:
Yes, that's just the reason I didn't use it lol.

rohit-kumar-j:
😆 Let me see if there is another method. Like an callback to this

Civitasv:
https://github.com/Civitasv/cmake-tools.nvim/blob/ba1a6cf3f0d6a27ed75c87926153907c1d0de1d6/lua/cmake-tools/utils.lua#L70-L80

I think this should be enough.

Civitasv:
First, we cd to the launch path, then we launch the executable.

Civitasv:
image
And I've tested it use this program, it says Success.

rohit-kumar-j:
> https://github.com/Civitasv/cmake-tools.nvim/blob/ba1a6cf3f0d6a27ed75c87926153907c1d0de1d6/lua/cmake-tools/utils.lua#L70-L80
> 
> I think this should be enough.

image

rohit-kumar-j:
pleanary job control has a [writer](https://github.com/nvim-lua/plenary.nvim/blob/9ac3e9541bbabd9d73663d757e4fe48a675bb054/lua/plenary/job.lua#LL18C56-L18C56) option. 

Seems like it is not exposed in the `:h plenary.nvim-pleanry-job`

Civitasv:
I think the reason for this fault is `vim.cmd(set_bufname)` ?

rohit-kumar-j:
Seems like it.

How about, something like this:  append to buffer if open?

Civitasv:
> How about, something like this: append to buffer if open?

I think that will do, do you know how to do it?

rohit-kumar-j:
Not sure yet. There is a pleanry [writer option](https://github.com/nvim-lua/plenary.nvim/blob/9ac3e9541bbabd9d73663d757e4fe48a675bb054/lua/plenary/job.lua#LL18C56-L18C56) that is not exposed in `:h plenary.nvim-pleanry-job` api. Perhaps this can solve it.

Civitasv:
> Opens mulitple terminals, that are not shown in the bufferline

Why is this happen? I don't quite understand.

rohit-kumar-j:
Seems to be opening multiple buffers as well:
(perhaps behaviour is undefined)

image

rohit-kumar-j:
previously it didn't show them in bufferline.

Civitasv:
Interesting 🧐

rohit-kumar-j:
**writer** seems to work with the api, but it isn't waiting for input(ofcourse I hardcoded it in):

```lua
  utils.job = Job:new({
    command = executable,
    args = {},
    cwd = opts.cmake_launch_path,
    writer = "1",
    on_stdout = vim.schedule_wrap(append_to_cmake_console),
    on_stderr = vim.schedule_wrap(append_to_cmake_console),
    on_exit = vim.schedule_wrap(function(_, code, signal)
      append_to_cmake_console("Exited with code " .. (signal == 0 and code or 118 + signal))
      if code == 0 and signal == 0 then
        if opts.on_success then
          opts.on_success()
        end
      elseif opts.cmake_show_console == "only_on_error" then
        utils.show_cmake_console(opts.cmake_console_position, opts.cmake_console_size)
        vim.api.nvim_command("cbottom")
      end
    end),
  })
```

image

rohit-kumar-j:
This works. Seems like as long as it is a table, it works:
```lua
function temp()
    return {"1"}
end
```
```lua
writer = temp(),
```

Civitasv:
But can it read read from standard input?

rohit-kumar-j:
need to return a stdin to writer in that case. Do you know how to create a function that can return a stdin?

Civitasv:
return io.read?

rohit-kumar-j:
ooh no.. terminal freezes.

Civitasv:
Maybe this api is not supposed to use like that.

rohit-kumar-j:
Yes seems like it does not have full support yet. Will try to do this using native `jobcontrol()` api.

rohit-kumar-j:
@Civitasv, this [vi stackexchange post](https://vi.stackexchange.com/questions/17816/solved-ish-neovim-dont-close-terminal-buffer-after-process-exit) works for not removing output from the terminal buffer.

Using the command `<C-\><C-n>` puts it in inspect mode as read only. Can this be an autocmd when process exits with code xyz?

Civitasv:
Have you seen the effect for pressing <C-\><C-n> on your computer?

rohit-kumar-j:
It shows `--TERMINAL--` in the statusline. No output until in the statusline until full sequence `<C-\><C-n>` and then no output.

Civitasv:
The terminal still disappeared?

rohit-kumar-j:
It does not disappear after `<C-\><C-n>`

However the setbufname is still giving error. Perhaps a pcall(), if something like that exists.

https://github.com/Civitasv/cmake-tools.nvim/assets/37873142/906c5f79-9dae-42dd-a8dc-6801148b35b1

Civitasv:
I wonder if it will be okay if we remove vim.cmd(set_bufname)

rohit-kumar-j:
if we purely do that, then unless we explicitly close the terminal buffer , it does not close the terminals(with `<C-\><C-n>`).
image

Civitasv:
> if we purely do that, then unless we explicitly close the terminal buffer , it does not close the terminals(with `<C-\><C-n>`).
> image

I thinks it should behave like this. And it will disappear if you first enter insert mode and then press any key, am i right?

rohit-kumar-j:
yes

rohit-kumar-j:
Something like this in `init.lua` for persistent output:

```lua
vim.api.nvim_create_autocmd("TermClose", { -- TermClose on proccess end
    group    = vim.api.nvim_create_augroup("cmaketools", { clear = true }),
    callback =  function()
        vim.cmd('call feedkeys("<C-\\><C-n>")')
    end
})
```

rohit-kumar-j:
> Something like this in `init.lua` for persistent output:
>
>```lua
>vim.api.nvim_create_autocmd("TermClose", { -- TermClose on proccess end
>    group    = vim.api.nvim_create_augroup("cmaketools", { clear = true }),
>    callback =  function()
>        vim.cmd('call feedkeys("<C-\\><C-n>")')
>    end
>})
>```

But this does not work: `vim.cmd('call feedkeys("<C-\\><C-n>")')`

Civitasv:
https://github.com/Civitasv/cmake-tools.nvim/blob/ba1a6cf3f0d6a27ed75c87926153907c1d0de1d6/lua/cmake-tools/utils.lua#L80
Actually, I have used this to start insert mode in terminal, maybe we can remove this line too?

rohit-kumar-j:
No, the `startinsert` is useful as it puts in insert mode immediately in insert ready for cin. Even if we remove cin, it still quits.

Civitasv:
> it still quits.

Weird, I cannot produce this situation, it wasn't quit in my computer.

rohit-kumar-j:
I now works! within `init.lua`:
```lua
vim.api.nvim_create_autocmd("TermClose", {
    group    = "cmaketools",
    callback =  function()
        vim.cmd.stopinsert()
        vim.api.nvim_feedkeys('<C-\\><C-n><CR>', 'n', false)
    end
})
```
maybe I can put a `if is_win32 then <create autocommand> end`

rohit-kumar-j:
Or make it optional form the config like `config.experimental = true`

Civitasv:
I think you can use `if is_win32`. No need to add it into configuration I think.

rohit-kumar-j:
How about this, check what bufs are avaliable in the instance and then delete if the buffer exists before spawnning `vim.cmd()`:

```lua
  local all_buffs = vim.api.nvim_list_bufs()
  local temp = " "
  for i, buf_nr in ipairs(all_buffs) do
    local name = vim.api.nvim_buf_get_name(buf_nr)
    local test = string.match(name, opts.cmake_launch_path) == opts.cmake_launch_path
    -- print(test)
    temp = temp .. name .. tostring(test) .. ", "
  end
  print(temp)

  ...
  vim.cmd(prefix ....)
  vim.bo.buflisted = true

  ...
```

It outputs even if it is alive:

First time:

image

Second time:

image

Civitasv:
I think it's feasible.

rohit-kumar-j:
Finally❗ (hope it works as well on linux)

https://github.com/Civitasv/cmake-tools.nvim/assets/37873142/25b31121-9117-48cc-9ec9-366ec197c14b

The buffers seem managed:
image

The output of `:lua vim.pretty_print(vim.api.nvim_list_bufs())` does not increase.

rohit-kumar-j:
I will send a draft PR after a little cleanup

Civitasv:
Good!

PR #53, opened 2023-05-19
Added CMakeLaunchArgs to set command line arguments

hfn92:
To allow for the launch target to be run with command line arguments I added `CMakeLaunchArgs`. This sets the the arguments to be passed to the currently selected launch target. Works for both `CMakeRun` and `CMakeDebug`. Tested only for linux.


Civitasv:
Will it be better if we parse the args passed to `CMakeRun` directly instead of making a standalone variable?

Civitasv:
But since we use `opt` also for `cmake.generate` if we have not generated and for `cmake.build` if we have not built, I don't know if it is feasible.

hfn92:
> Will it be better if we parse the args passed to `CMakeRun` directly instead of making a standalone variable?

> But since we use `opt` also for `cmake.generate` if we have not generated and for `cmake.build` if we have not built, I don't know if it is feasible.

Yeah I also thought about doing that first. But I found this solution to be more convenient since it remembers the arguments for each target individually and they are used for both `CMakeRun`  and `CMakeDebug` . I often have multiple executables that require different arguments.

Civitasv:
That makes sense.

PR #54, opened 2023-05-19
build_directory ignored when debugging

hfn92:
When debugging the current working directory is set, whereas when running normally the build directory is used


Civitasv:
@rohit-kumar-j Can you review it?

rohit-kumar-j:
Unfortunately I donot havr dap setup, otherwise it seems alright.

Civitasv:
Okay. It should be okay.

PR #56, opened 2023-05-24
fix: preset source dir and generator;

LinHeLurking:
This PR fix 2 broken things when generating with CMakePresets.json

1. Wrong binary dir
2. Not recognizing generator


LinHeLurking:
Sorry! It seems that different version of cmake behave differently. It's not this plugin's fault not correctly recognizing `binaryDir` or `generator`. I'll survey more and give feedback.

Issue #55, opened 2023-05-21
Persistent output buffers


rohit-kumar-j:
I have been working on a PR for two output buffers. One for all cmake tasks except run, and another exclusively for run/launch tasks. 

The intention for this is to keep track of outputs from all the previous commands that cmake tools generates and provide an additonal way to view the outputs not just with quickfix lists.(... As nature intended.. XD) 

I have setup the window/buffer handling in way that launches a buffer first (if not already exists) that has a unique id that persists across the neovim instance.

The following shortened piece of code is intended to launch a job (using native vim/neovim api) and append to an exisint buffer. Here, the problem is in sending lines to the job with stdin. I would appreciate it, if someone can help on this:

Kindly run it and see the output:

```lua
local function on_stdout(id, data, event) vim.fn.append(vim.fn.line('$') - 1, data) end
local job_id = vim.fn.jobstart({ "pwsh", "-noprofile" }, { on_stdout = on_stdout })

local function callback(input) vim.fn.chansend(job_id, { input, "" }) end

local function goto_buffer_across_all_tabs(bufname)
    for _, tabpage in ipairs(vim.api.nvim_list_tabpages()) do
        vim.api.nvim_set_current_tabpage(tabpage)
        for _, window in ipairs(vim.api.nvim_tabpage_list_wins(tabpage)) do
            local win_bufname = vim.api.nvim_buf_get_name(vim.api.nvim_win_get_buf(window))
            if win_bufname:match(bufname) then
                vim.api.nvim_set_current_win(window)
                return
            end
        end
    end
end

-- vim.cmd([[command! -nargs=* WinGo2 lua goto_buffer_across_all_tabs(<q-args>)]])

local bufnr = vim.api.nvim_create_buf(true, true)
vim.api.nvim_buf_set_name(bufnr, "new")
vim.cmd('vs' .. bufnr)
goto_buffer_across_all_tabs('new')

vim.bo.buftype = "prompt"
vim.api.nvim_get_current_buf()
vim.api.nvim_set_current_buf(bufnr)
vim.fn.prompt_setcallback(bufnr, callback)
vim.keymap.set("n", "<esc>", ":bw!<cr>")
vim.cmd.startinsert()
```

For testing in unix change pwsh to /bin/sh

Currently this is the output:

![image.png](https://github.com/Civitasv/cmake-tools.nvim/assets/37873142/857b4d37-a0e8-44a7-9e48-674a2195b3b0)

[...]

PR #57, opened 2023-05-26
Working with terminals

rohit-kumar-j:
## This is a W.I.P.

I am opening this dirty draft PR, as a communication channel. Here is what I am currently doing:

1. The `on_success` callbacks are not implemented (require pipes?)
2. The `os_config.lua` file is only for dealing with child procs and their outputs. (If there is a suitable method where the output from the buffer does not get overwritten by the terminal emulator, this method would be ideal.

## TODO:

- [ ] vim.schedule/wrap the creation of terminals
- [ ] on_success callback command to the terminal (Either child procs or via pipes)
- [ ] Implement `cmake.stop()` for terminal
- [ ] Implement `cmake.install()` for terminal
- [ ] Add window handling back(which is the whole point of this PR)

Since my [inital test with window handling](https://github.com/Civitasv/cmake-tools.nvim/issues/55#issuecomment-1556581793), my local branch of cmake-tools was removed by [lazy](https://github.com/folke/lazy.nvim) 💀, since then I had to redo much of it, and so, currently there is no window handling.

### DIFF
--- a/lua/cmake-tools/config.lua
+++ b/lua/cmake-tools/config.lua
@@ -18,6 +18,7 @@ local Config = {
   kit = nil,
   configure_preset = nil,
   build_preset = nil,
+  cmake_startup = nil,
   main_terminal = nil,
   run_executable_terminals = {},
   debug_executable_terminals = {}
--- a/lua/cmake-tools/config.lua
+++ b/lua/cmake-tools/config.lua
@@ -18,6 +18,9 @@ local Config = {
   kit = nil,
   configure_preset = nil,
   build_preset = nil,
+  main_terminal = nil,
+  run_executable_terminals = {},
+  debug_executable_terminals = {}
 }
 
 function Config:new(const)
--- a/lua/cmake-tools/const.lua
+++ b/lua/cmake-tools/const.lua
@@ -27,6 +27,8 @@ local const = {
     keep_single_terminal_static = true,
     only_1_terminal_per_tab = true,
     single_terminal_window_per_tab = true,
+    terminal_split_direction = 'below',
+    terminal_split_size = 15,
   }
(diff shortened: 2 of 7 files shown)


rohit-kumar-j:
@Civitasv, could you kindly have a look at this, from the perspective of sending commands to the buffers? 
I will edit and add the windowing functionality back to this in the process.

rohit-kumar-j:
Generate, Clean, build work individualy, but not when chained.(currently using only 1 main buffer)

This is how it looks:
image

rohit-kumar-j:
For refernce for creating similar unix commands:
Wrapping the smiple commands in a child process for running them in the Powershell:

run `:term` first. after that:

```lua
:lua vim.api.nvim_chan_send(vim.api.nvim_buf_get_var(vim.api.nvim_win_get_buf(vim.api.nvim_get_current_win()),"terminal_job_id"), "Start-Process -FilePath pwsh -ArgumentList '-Command Start-Sleep -Seconds 5 && ls && echo \"done!\" ' -PassThru -NoNewWindow | Wait-Process \r"))
```

Before the timeout completes:
```lua
:lua vim.print(vim.api.nvim_get_proc_children(vim.api.nvim_buf_get_var(vim.api.nvim_win_get_buf(vim.api.nvim_get_current_buf()),"terminal_job_pid")))
```

### Notice: "terminal_job_id" vs "terminal_job_pid" in the commands
### We can also send (") with: echo \\"done!\\"

<details>
<Summary>How it looks</Summary>

image

</details>

Reference : https://github.com/neovim/neovim/issues/23777#issuecomment-1565397726

rohit-kumar-j:
Creating new PR

PR #58, opened 2023-05-28
Multi Terminals and Persistent buffers

rohit-kumar-j:
## TODO:

- [ ] Process wrapping in linux (in utils.lua)
- [ ] Process wrapping in macos (in utils.lua)
- [ ] Multi terminal handling


rohit-kumar-j:
@Civitasv, a basic demo is ready (currently without window handling)... can you please take a look?

Civitasv:
I'm back to work, sorry for my late. Testing now.

Civitasv:
In order to use this feature, I need to turn on cmake_use_terminals, am i right?

Civitasv:
test
I modified a little for linux, now it works like this, is it expected behaviour?

rohit-kumar-j:
> In order to use this feature, I need to turn on cmake_use_terminals, am i right?

Yes

As of now,there is no window handling. You can test all other settings in  `cmake_terminal_opts` except:
```lua
    -- Window handling
    display_single_terminal_arcoss_instance = true,
    single_terminal_pet_tab = true,
    keep_terminal_in_static_location = true,

```

> I modified a little for linux, now it works like this, is it expected behaviour?

It should resize the window to 10(height), by default... (unless you have explicitly set ti higher)

Civitasv:
I don't know why do you use `launch_task_in_a_child_process`.

rohit-kumar-j:
> I don't know why do you use `launch_task_in_a_child_process`.

Haven't added multterminal support(Each executable target has it's own terinal + 1 Main Terminal) yet. 
So I've kept it for now.

Civitasv:
Okay, I understand.

Civitasv:
> * Fix first time `:CMakeGenerate` bug: ('E21: modifiable is off', does not prompt the user for build type after asking for the desired kit)

I cannot reproduce this problem.

Civitasv:
I've added support for Linux.

rohit-kumar-j:
> > * Fix first time `:CMakeGenerate` bug: ('E21: modifiable is off', does not prompt the user for build type after asking for the desired kit)
> 
> I cannot reproduce this problem.

It does not implement the `on_success` method in any way ...yet... (Not sure how to implement it)

rohit-kumar-j:
Delete the build folder and try doing a `:CMakeRun` directly without a `:CMakeGenerate` and `:CMakeBuild`. 
The commands do not chain with any `on_success()` methods 😞 i.e. it first generates, then `on_success()`, it builds, then `on_success()`, it runs.(each time asking for options). 

Does this happen for you?

rohit-kumar-j:
> Delete the build folder and try doing a `:CMakeRun` directly without a `:CMakeGenerate` and `:CMakeBuild`. The commands do not chain with any `on_success()` methods 😞 i.e. it first generates, then `on_success()`, it builds, then `on_success()`, it runs.(each time asking for options).
> 
> Does this happen for you?

I think I will ask for a all the options from the user first(like `Kit` and `Build Type`), then do a sequentail chansend. This will be the best option, but a little tricky to implement, especially within child processes. It may be a workaround for chaining commands.  

What do you think of this?

rohit-kumar-j:
@Civitasv, can you add support for linux in `utils.send_data_to_terminal()`?

Civitasv:
I've extracted the terminal logic to `terminal.lua`, and I think it is not necessary to use terminal when generating and building, but I still keep those codes in `terminal.lua`(in `terminal.run` function). Now `utils.execute` will invoke `terminal.execute`, which will send command to our unified terminal. What do you think?

Peek 2023-05-30 12-03

rohit-kumar-j:
Interesting! This can be an option as well. 

The purpose was persistent output, but some users might not want persistent output for tasks other than run/debug, they can choose to run with terminal and build with quickfix lists.

Civitasv:
> Interesting! This can be an option as well.
> 
> The purpose was persistent output, but some users might not want persistent output for tasks other than run/debug, they can choose to run with terminal and build with quickfix lists.

Yes

rohit-kumar-j:
> Delete the build folder and try doing a `:CMakeRun` directly without a `:CMakeGenerate` and `:CMakeBuild`. The commands do not chain with any `on_success()` methods 😞 i.e. it first generates, then `on_success()`, it builds, then `on_success()`, it runs.(each time asking for options).
> 
> Does this happen for you?

solved for terminal mode. with `vim.schedule_wrap()`

rohit-kumar-j:
For the qf list, this is the erro I get:

```bash
|| CMake Error at CMakeLists.txt:3 (project):
||   The CMAKE_CXX_COMPILER:
|| 
||     ;C:/Program;Files/LLVM/bin/clang.exe""
|| 
||   is not a full path and was not found in the PATH.
```
It does not select the correct compiler strings. The problem is in `kits.build_env_and_args()`

It should be `"C:/Program Files/LLVM/bin/clang.exe"`

rohit-kumar-j:
Otherwise, it seems good to go, at-least for the terminals.

Civitasv:
> Do not need on success as it is all in the terminal. This will chain the commands using `chan_send()` to the terminal anyways.

Can you explain more? I don't get it worked on Linux.

Civitasv:
Peek 2023-05-31 10-34

Do you know the reason that it doesn't work as expected? It doesn't chain the commands. But I wonder if we don't invoke `on_success` callback, how could it run other chained command?

Or Maybe I did wrong in `terminal.send_data_to_terminal`?

Civitasv:
Or maybe you forget to add on_success call in terminal.run?

rohit-kumar-j:
> > Do not need on success as it is all in the terminal. This will chain the commands using `chan_send()` to the terminal anyways.
> 
> Can you explain more? I don't get it worked on Linux.

Yes, when we do `:CMakeGenerate` or `:CMakeBuild` or `:CMakeRun`, the compiler, build type or launch target are not selected, if we do this:
```lua
vim.ui.select(types, { prompt = "Select build type" }, callback)
```
,instead of this:
```lua
vim.ui.select(types, { prompt = "Select build type" }, vim.schedule_wrap(callback))
```

In the functions:
```lua
cmake.select_build_type(callback)
cmake.select_kit(callback)
cmake.select_configure_preset(callback)
cmake.select_build_preset(callback)
cmake.select_build_target(callback)
cmake.select_launch_target(callback)
```
to:
```lua
cmake.select_build_type(vim.schedulw_wrap(callback))
cmake.select_kit(vim.schedulw_wrap(callback))
cmake.select_configure_preset(vim.schedulw_wrap(callback))
cmake.select_build_preset(vim.schedulw_wrap(callback))
cmake.select_build_target(vim.schedulw_wrap(callback))
cmake.select_launch_target(vim.schedulw_wrap(callback))
```

With `vim.schedule_wrap()`, the `vim.ui.select()` works well with both Quickfix Lists and Terminals

lua/cmake-tools/init.lua
```diff
@@ -352,38 +353,55 @@ function cmake.run(opt, callback)
...
-            cmake_launch_args = cmake:get_launch_args(),
-            cmake_unify_terminal_for_launch = const.cmake_unify_terminal_for_launch,
-            cmake_terminal_opts = const.cmake_terminal_opts
-          })
-        else
-          if not const.cmake_unify_terminal_for_launch then
-            target_path = '"' .. target_path .. '"'
+    if const.cmake_unify_terminal_for_launch then
+      -- DONOT schedule (or) schedule_wrap this piece of code
+      -- BUG: Without this if-else logic, the forwarding to terminal.execute() from utils.execute(), in utils.lua does not work.
```
rohit-kumar-j, line 358: @Civitasv, please do not remove this if-else logic, as there is a bug on my machine. It does not call `terminal.execute()` from `utils.execute()`.

I will remove it later, but for now works. Otherwise, it launches in a quickfix list from `utils.run()` instead of `utils.execute()`...

Civitasv:
> 

I know why you use `vim.schedule_wrap`, but it still doesn't work on Linux.

Civitasv:
```lua
    if const.cmake_unify_terminal_for_launch then
      -- DONOT schedule (or) schedule_wrap this piece of code
      -- BUG: Without this if-else logic, the forwarding to terminal.execute() from utils.execute(), in utils.lua does not work.
      result = config:get_launch_target()
      local target_path = result.data
      local new_s = vim.fn.fnamemodify(target_path, ":h")
      return terminal.execute(target_path, {
        bufname = vim.fn.expand("%:p"),
        cmake_launch_path = new_s,
        cmake_console_position = const.cmake_console_position,
        cmake_console_size = const.cmake_console_size,
        cmake_launch_args = cmake:get_launch_args(),
        cmake_use_terminals = const.cmake_unify_terminal_for_launch,
        cmake_terminal_opts = const.cmake_terminal_opts
      })
    else
```

And I noticed you've removed `cmake.build` wrapper, I don't think this will work. We should build before execute. Like:

```lua
      return cmake.build({ fargs = utils.deepcopy(opt.fargs) }, function()
        vim.schedule(function()
          result = config:get_launch_target()
          local target_path = result.data
          local new_s = vim.fn.fnamemodify(target_path, ":h")

          if (osys.is_win32 == 1) then
            return utils.execute(target_path, {
              bufname = vim.fn.expand("%:p"),
              cmake_launch_path = new_s,
              cmake_console_position = const.cmake_console_position,
              cmake_console_size = const.cmake_console_size,
              cmake_launch_args = cmake:get_launch_args(),
              cmake_unify_terminal_for_launch = const.cmake_unify_terminal_for_launch,
              cmake_terminal_opts = const.cmake_terminal_opts
            })
          else
...
```

Civitasv:
I wonder if you can install a Linux distro and test it your own, that may make things easier, On Windows, you can use WSL.

lua/cmake-tools/init.lua
Civitasv: Indeed weird.

rohit-kumar-j:
> And I noticed you've removed cmake.build wrapper, I don't think this will work. We should build before execute. Like:

Ah yes. I do not remember why I did this. Will add it back after window handling.

rohit-kumar-j:
@Civitasv, can you check if all the options are working as expected on linux?

Civitasv:
When `cmake_use_terminal_for_build` is set to `true`, then invoke `CMakeRun`, it will show error:

You must build the executable first!... Use ":CMakeBuild"

rohit-kumar-j:
Unfortunately, If we remove that, this is the error:

image

The `"v:"` should be `"cd /<path to executable>"` and `./null` should be `.\executable_name.exe`

I will try and find a strategy to mitigate this, although, I'm not sure if it is possible to.

rohit-kumar-j:
SInce there is no `on_success()` method for passing a command directly to the termial. 

So, we cannot do a `:CMakeRun` after a `:CMakeBuild`, which itself would be ran after a `:CMakeGenerate`, directly if we do `:CMakeRun` on statup. It wil generate, and not build or run. As the `on_success()` method requires wrapping commands inside a child process.

rohit-kumar-j:
So, we have to do a `:CMakeGenerate` first, then a `:CMakeBuild` and finally a `:CMakeRun` individuallly if we use a terminals for all other tasks.

rohit-kumar-j:
Can you have a look at the `cmake.select_build_target()` and `cmake.select_launch_target()` logic? I think the issue is there. If we can solve it, we can remove this `check if executable==nil`.

Civitasv:
> So, we have to do a `:CMakeGenerate` first, then a `:CMakeBuild` and finally a `:CMakeRun` individuallly if we use a terminals for all other tasks.

I don't think it is usable, it may be better to use quicklist for `CMakeGenerate`, `CMakeBuild`, then only use terminal for `CMakeRun`.

rohit-kumar-j:
Wouldn't it be alright if we tell users that using the terminals for build tasks would not chain commands and they need to use it with caution?

Civitasv:
> Wouldn't it be alright if we tell users that using the terminals for build tasks would not chain commands and they need to use it with caution?

If so, I don't think any user will use this method. Cause It is far more convenient to chain all commands.

Civitasv:
I've extracted some logic about quickfix from utils.lua to quickfix.lua, and I've changed some options' name, to make code clearer.

And I wonder what's the meaning of option `keep_terminal_static_location`?

rohit-kumar-j:
>> Wouldn't it be alright if we tell users that using the terminals for build tasks would not chain commands and they need to use it with caution?
>
> If so, I don't think any user will use this method. Cause It is far more convenient to chain all commands.

For example, if you want to pass some args manually. I was working with Clang on windows and had to do [this](https://metricpanda.com/rival-fortress-update-27-compiling-with-clang-on-windows/) and [this](https://stackoverflow.com/questions/46553436/building-with-cmake-ninja-and-clang-on-windows) which required manual tweaking of the cmake commands and setting the environments for using clang and ninja.

rohit-kumar-j:
> And I wonder what's the meaning of option `keep_terminal_static_location`?

The logic currently works like this... if you set `single_terminal_per_instance` to true and `keep_terminal_static_location` to false, then terminal will close and popup below the currently active window. We skip this closing the terminal part if we set `keep_terminal_static_location` to true.

rohit-kumar-j:
Nothing is happening with `cmake_always_use_terminal` on my machine

Civitasv:
> > And I wonder what's the meaning of option `keep_terminal_static_location`?
> 
> The logic currently works like this... if you set `single_terminal_per_instance` to true and `keep_terminal_static_location` to false, then terminal will close and popup below the currently active window. We skip this closing the terminal part if we set `keep_terminal_static_location` to true.

Then when do we prefer it set to true and when do we prefer it set to false, in my understanding, i will always set it to true.

Civitasv:
> Nothing is happening with `cmake_always_use_terminal` on my machine

It is a rename from cmake_use_terminal_for_build.

rohit-kumar-j:
I think quickfix.lua and utils.lua seperation is alright, but we should keep the `cmake_use_terminal_for_build` and `cmake_unify_terminal_for_launch`, as you had mentioned eralier if the user wants to use quick_fix for tasks and terminal for launch.

rohit-kumar-j:
> > > And I wonder what's the meaning of option `keep_terminal_static_location`?
> > 
> > 
> > The logic currently works like this... if you set `single_terminal_per_instance` to true and `keep_terminal_static_location` to false, then terminal will close and popup below the currently active window. We skip this closing the terminal part if we set `keep_terminal_static_location` to true.
> 
> Then when do we prefer it set to true and when do we prefer it set to false, in my understanding, i will always set it to true.

We set it to true if we have mutliple splits and want to spawn the terminal the below currently focused split.

Civitasv:
> I think quickfix.lua and utils.lua seperation is alright, but we should keep the `cmake_use_terminal_for_build` and `cmake_unify_terminal_for_launch`, as you had mentioned eralier if the user wants to use quick_fix for tasks and terminal for launch.

Currently, the idea is always use terminal for launch, and if cmake_always_use_terminal is set to true, we also use terminal for generate, build, and etc, else we use quickfix for generate, build and etc.

rohit-kumar-j:
> > > Wouldn't it be alright if we tell users that using the terminals for build tasks would not chain commands and they need to use it with caution?
> > 
> > 
> > If so, I don't think any user will use this method. Cause It is far more convenient to chain all commands.
> 
> For example, if you want to pass some args manually. I was working with Clang on windows and had to do [this](https://metricpanda.com/rival-fortress-update-27-compiling-with-clang-on-windows/) and [this](https://stackoverflow.com/questions/46553436/building-with-cmake-ninja-and-clang-on-windows) which required manual tweaking of the cmake commands and setting the environments for using clang and ninja.

And in general, the actual commands are seen as if we passed it directly into the terminal... which is similar to VS Code's implementation where we use `CMake:Show Configure Command` or `CMake:Show Launch Command` to see the full that is command passed. (And may even be better as we can tweak the command's args)

rohit-kumar-j:
> > I think quickfix.lua and utils.lua seperation is alright, but we should keep the `cmake_use_terminal_for_build` and `cmake_unify_terminal_for_launch`, as you had mentioned eralier if the user wants to use quick_fix for tasks and terminal for launch.
> 
> Currently, the idea is always use terminal for launch, and if cmake_always_use_terminal is set to true, we also use terminal for generate, build, and etc, else we use quickfix for generate, build and etc.

Oh my bad. I was under the impression that we can either use quickfix for simple launch commands or the terminal.

rohit-kumar-j:
> > Yes, when set it to true, it is a little less disorienting.
> 
> Then, maybe we just make it as default, no need to expose an option.

Cool. I will keep it as an internal option as of now. I think it might be useful for debugging split handling.

Civitasv:
> For example, if you want to pass some args manually. I was working with Clang on windows and had to do [this](https://metricpanda.com/rival-fortress-update-27-compiling-with-clang-on-windows/) and [this](https://stackoverflow.com/questions/46553436/building-with-cmake-ninja-and-clang-on-windows) which required manual tweaking of the cmake commands and setting the environments for using clang and ninja.

You can add these options after CMakeGenerate or CMakeBuild, like CMakeGenerate -Da=b

Civitasv:
But I agree with you that having a unified terminal version is good. I will try to find a suitable way to chain commands. Otherwise, i think we should not publish this feature.

rohit-kumar-j:
> > For example, if you want to pass some args manually. I was working with Clang on windows and had to do [this](https://metricpanda.com/rival-fortress-update-27-compiling-with-clang-on-windows/) and [this](https://stackoverflow.com/questions/46553436/building-with-cmake-ninja-and-clang-on-windows) which required manual tweaking of the cmake commands and setting the environments for using clang and ninja.
> 
> You can add these options after CMakeGenerate or CMakeBuild, like CMakeGenerate -Da=b

In the vim commands? Like `:CMakeGenerate -D... = ...`?

Civitasv:
Yes.

rohit-kumar-j:
Nice. I didn't know this.😖

rohit-kumar-j:
> But I agree with you that having a unified terminal version is good. I will try to find a suitable way to chain commands. Otherwise, i think we should not publish this feature.

This may be cumbersome, but wrapping them in child process does the trick. We can do something like: 

```lua
local cprocs = nvim_get_proc_children()
escape_loop_flag = false

while(escape_loop_flag == false) do
    if cprocs==0 then
    -- New command : on_sucess()
    -- set escape_loop_flag = true
    end
end
```
But we need to do this is async and wrap in child processes. This is the best I have right now.

rohit-kumar-j:
Is there a `cmake_generate_options` per project?

Civitasv:
> Is there a `cmake_generate_options` per project?

That’s what cmake kits and cmake presets do.

Civitasv:
> > But I agree with you that having a unified terminal version is good. I will try to find a suitable way to chain commands. Otherwise, i think we should not publish this feature.
> 
> This may be cumbersome, but wrapping them in child process does the trick. We can do something like:
> 
> ```lua
> local cprocs = nvim_get_proc_children()
> escape_loop_flag = false
> 
> while(escape_loop_flag == false) do
>     if cprocs==0 then
>     -- New command : on_sucess()
>     -- set escape_loop_flag = true
>     end
> end
> ```
> 
> But we need to do this is async and wrap in child processes. This is the best I have right now.

Okay. Have you tested it on Windows?

rohit-kumar-j:
> Okay. Have you tested it on Windows?

Yes.
But it is cumbersome:

Instead of `cmd` = `cmd` for 
`cmd` = `cmake -B build -S . ...`, It is:
`cmd` = `Start-Process -FilePath pwsh -ArgumentList '-Command "` .. cmd .. `" ' -PassThru -NoNewWindow | Wait-Process \r`

Civitasv:
> > Okay. Have you tested it on Windows?
> 
> Yes. But it is cumbersome:
> 
> Instead of `cmd` = `cmd` for `cmd` = `cmake -B build -S . ...`, It is: `cmd` = `Start-Process -FilePath pwsh -ArgumentList '-Command "` .. cmd .. `" ' -PassThru -NoNewWindow | Wait-Process \r`

Oh, i see that in terminal.lua. I will see what i can do in linux.

rohit-kumar-j:
> > > Okay. Have you tested it on Windows?
> > 
> > 
> > Yes. But it is cumbersome:
> > Instead of `cmd` = `cmd` for `cmd` = `cmake -B build -S . ...`, It is: `cmd` = `Start-Process -FilePath pwsh -ArgumentList '-Command "` .. cmd .. `" ' -PassThru -NoNewWindow | Wait-Process \r`
> 
> Oh, i see that in terminal.lua. I will see what i can do in linux.

I think it is something like `cmd` = `cmd` .. `& \r` for linux.

rohit-kumar-j:
I will revert back the **feat. extract logic for quickfix in quickfix.lua, and make code clearer,** commit and seperate the `quickfix.lua` and `utils.lua` as you have done, and prepare it for child procs. 

Something in thisc commit broke inidividual commands to terminal on windows.

Civitasv:
> Something in thisc commit broke inidividual commands to terminal on windows.

Break what? Maybe we should merge a version and make a new pull request. Too many changes here.

rohit-kumar-j:
> > Something in thisc commit broke inidividual commands to terminal on windows.
> 
> Break what? Maybe we should merge a version and make a new pull request. Too many changes here.

Sorry. something with neovim not updating my config or something. Do not know why it took so long, even when I `require('cmaketools')`. It works as expected.

Civitasv:
Weird, do you have ninja installed? What’s the full command?

rohit-kumar-j:
> Weird, do you have ninja installed? What’s the full command?

Ninja: Yes. Unfortunately, there is no full cmake command. Only this shows up in the QF list if I do generate.

rohit-kumar-j:
> > Weird, do you have ninja installed? What’s the full command?
> 
> Ninja: Yes. Unfortunately, there is no full cmake command. Only this shows up in the QF list if I do generate.

Pushed a fix. Can you review this [`613757c`](https://github.com/Civitasv/cmake-tools.nvim/pull/58/commits/613757c170235ea3e746c976dc7a3959683692d9)

rohit-kumar-j:
Is there a way to not show this output each time I generate?

image

Civitasv:
> Is there a way to not show this output each time I generate?
> 
> image

Maybe vim.cmd has an option to silently execute, I don’t know yet.

rohit-kumar-j:
> > Is there a way to not show this output each time I generate?
> > image
> 
> Maybe vim.cmd has an option to silently execute, I don’t know yet.

Maybe something like {silent=true} might do the trick. 

---

Anyways, this does not work `cmake_always_use_terminal = true` in windows. There is no output.

rohit-kumar-j:
The problem is here:
```lua
function utils.has_active_job(always_use_terminal)
  if always_use_terminal then
    return terminal.has_active_job()
  else
    return terminal.has_active_job() or quickfix.has_active_job()
  end
end
```

This only works with child procs, not directly sending commands to terminal

Civitasv:
Okay, how can we fix it?

rohit-kumar-j:
Pushed a fix. It also ensures that we can further develop directly with the terminal.

The chaining commands works if the project is configured first(As we can get the corresponding targets first, and then chain commands). 

Maybe run configure and env creation first in a QFList?

Civitasv:
> Maybe run configure and env creation first in a QFList?

okay.

Civitasv:
> Maybe vim.cmd has an option to silently execute, I don’t know yet.

fix.

Civitasv:
Why Reposition term is not supported for running task/executable child processes?

Civitasv:
`nvim_chan_send` is non-blocking

Civitasv:
I've made `cmake_always_use_terminal` work on Linux platform, with a little hack, but it works.

```lua
  -- Send final cmd to terminal
  terminal.send_data_to_terminal(buffer_idx, full_cmd,
    {
      wrap = opts.cmake_terminal_opts.launch_task_in_a_child_process,
      win_id = final_win_id,
      split_direction = opts.cmake_terminal_opts.split_direction,
      split_size = opts.cmake_terminal_opts.split_size,
      start_insert = opts.cmake_terminal_opts.start_insert_in_other_tasks,
      focus_on_main_terminal = opts.cmake_terminal_opts.focus_on_main_terminal,
    })

  vim.wait(10000000, function()
    if not terminal.has_active_job() then
      -- it should have an active job
      return false
    end
    return true
  end, 500)

  vim.wait(10000000, function()
    if terminal.has_active_job() then
      -- it should not have an active job
      return false
    end
    return true
  end, 500)

  if opts.on_success then
    opts.on_success()
  end
```

In this way, we can chain commands.

And, I think it's no need to specify `launch_task_in_a_child_process`, cause every command in shell is a child process for it. So `terminal.check_if_running_child_procs` works fine even `launch_task_in_a_child_process` is false. So I think it's no need to wrap cmd to child process, like what you write `Start-Process` on Windows. Can you test it on Windows, if it works, we can remove this option.

But, there is still a weird problem, at the first time we open a project, the command `CMakeBuild` won't be able to chain commands, but if we invoke it another time, it works fine then. I have no idea about the reason. Maybe you can give me some idea.

Civitasv:
And, what's the difference between `check_cmake_buffers_are_displayed_in_current_tab` and `check_if_cmake_buffers_are_displayed_in_current_tab`

Civitasv:
- [ ] Test on Windows.
- [ ] Big changes in README.
- [ ] Add comments for function.

rohit-kumar-j:
> > But, there is still a weird problem, at the first time we open a project, the command `CMakeBuild` won't be able to chain commands, but if we invoke it another time, it works fine then. I have no idea about the reason. Maybe you can give me some idea.
> 
> Same for CMakeGenerate.

Yes, this is probably beacuse the project is not configured the first time.

Civitasv:
> Yes, this is probably beacuse the project is not configured the first time.

No, i have blocked it till finished. So opts.on_success should work well.

The problem is: the first time I enter cmake project and invoke CMakeBuild, it will call CMakeGenerate correctly. And after finishing CMakeGenerate, it also correctly calls cmake.build, but this time, it cannot get build targets correctly.

After this failure, every time it works fine. Even after I delete build directory and call CMakeBuild, it chains commands correctly. So it’s really confusing. I don’t know if i make it clear. Maybe you can try yourself.

rohit-kumar-j:
> And, what's the difference between `check_cmake_buffers_are_displayed_in_current_tab` and `check_if_cmake_buffers_are_displayed_in_current_tab`

Sorry, wrong naming. Pushed a fix. [`700c450`](https://github.com/Civitasv/cmake-tools.nvim/pull/58/commits/700c450e5671f8e5bdfcd832dc525a4bb4ef1648)

rohit-kumar-j:
> The problem is: the first time I enter cmake project and invoke CMakeBuild, it will call CMakeGenerate correctly. And after finishing CMakeGenerate, it also correctly calls cmake.build, but this time, it cannot get build targets correctly.

What is the config for this? 
Do you use   `cmake_always_use_terminal = true` for this?

rohit-kumar-j:
I'm getting this error on a configured project:
image

Civitasv:
Yes.

rohit-kumar-j:
### First Try

When I do a `:CMakeRun` on new project(unconfigured, no build directory), it chains the following:
1. `:CMakeGenerate`
2. `:CMakeClean`
3. `cmake -E create_symlink ...json ...json `
4. `cmake -E create_symlink ...json ...json `

Config: 
`cmake_always_use_terminal = true`,

---

### Second Try

When I do a `:CMakeRun` on a configured and built project, but Vim Just Entered (`:h VimEnter`), it detects the project's targets, but does not create a new terminal with the incorrect name:
`[CMakeTools]: null`. It does not eagerly detect the names of the targets.

Config: 
`cmake_always_use_terminal = true`,

rohit-kumar-j:
I have an idea. 

For a new project, for the first time, when we do a `:CMakeGenerate`, it configures the project. 
After configuring, only then we get a `CMakeCache.txt`. So, we can have a loop that detects if a `CMakeCache.txt` file is created in the build directory. Only after this it is configured, we can use the `codemodel targets`.

If this is done in async, I think there might not be too many calls to the filesystem.

Although this is a CMake specific thing, I tested this for both MSVC(VS Generator) and Clang(Ninja). The `CMakeCache.txt` seems to be generated only after the configure step is complete.

Civitasv:
> I encountered (a similar/the same?) problem...

Kinda same. But for your second try, it works correctly on Linux.

Civitasv:
> I have an idea.
> 
> For a new project, for the first time, when we do a `:CMakeGenerate`, it configures the project. After configuring, only then we get a `CMakeCache.txt`. So, we can have a loop that detects if a `CMakeCache.txt` file is created in the build directory. Only after this it is configured, we can use the `codemodel targets`.
> 
> If this is done in async, I think there might not be too many calls to the filesystem, so as to block the IO. If we get a hit on the `CMakeCache.txt`, we can raise a flag and end the loop. This triggers the `get_targets...()` functions.
> 
> Although this is a CMake specific thing, I tested this for both MSVC(VS Generator) and Clang(Ninja). The `CMakeCache.txt` seems to be generated only after the configure step is complete.

Yeah, that’s a way to fix it. But it is really weird.

rohit-kumar-j:
Before doing this, I think we can do some clean up and merge PR. 😅

Can you add the description of the terminal options in the README? 
I'm not sure what to write. There was so much in this PR.

Civitasv:
I think launch_task_in_a_child_process and launch_executable_in_a_child_process can be removed.

Civitasv:
I think launch_task_in_a_child_process and launch_executable_in_a_child_process can be removed.

rohit-kumar-j:
I agree.

rohit-kumar-j:
Let's keep it internally?

Civitasv:
Also remember remove the logic about child procs in terminal.lua

Civitasv:
> Let's keep it internally?

Don’t think it will be useful someday.

rohit-kumar-j:
Completely remove them or store them in a .legacy.lua file that isn't required by any of the other files?

rohit-kumar-j:
> > Let's keep it internally?
> 
> Don’t think it will be useful someday.

Okay

Civitasv:
> Before doing this, I think we can do some clean up and merge PR. sweat_smile
> 
> Can you add the description of the terminal options in the README? I'm not sure what to write. There was so much in this PR.

I will add it today.

rohit-kumar-j:
There seems to be problem with

```lua 
  vim.wait(10000000, function()
    if not terminal.has_active_job() then
      -- it should have an active job
      return false
    end
    return true
  end, 10)

  vim.wait(10000000, function()
    if terminal.has_active_job() then
      -- it should not have an active job
      return false
    end
    return true
  end, 10)
```

I have to do `CTRL C` twice for seeing any output from neovim. It is blocking even the viewport.

Civitasv:
Yes, I don't know hot to make it async yet.

rohit-kumar-j:
> Yes, I don't know hot to make it async yet.

Although [not fully async](https://www.youtube.com/watch?v=GToxQKi7LZg&ab_channel=TJDeVries) according to TJ, this can be a [starting point](https://www.youtube.com/watch?v=W8Mq--dqNow&ab_channel=YukiUthman).

I'm trying to use this.

Civitasv:
Okay... Maybe we should find another way. I will update README first.

Civitasv:
Make a new issue, we can continue discuss there. #60

PR #62, opened 2023-06-10
Stop saving twice?

rohit-kumar-j:
Do not save in execute, as we use terminals... if commands are chained, it just saves twice. The print message twice is ugly as we need to hit enter to escape the print message.


Civitasv:
I think it's better to make this cmd silent.

Civitasv:
vim.cmd("silent exec " .. "\"wall\""

Civitasv:
> `vim.cmd("silent exec " .. "\"wall\"")`

I've changed to this command.

Issue #60, opened 2023-06-06
Terminal mode: When checking if there is a process attached to terminal, it should not block neovim instance


Civitasv:
Currently, I use these codes to check if there is a process is finished, which is need to chain callback.

https://github.com/Civitasv/cmake-tools.nvim/blob/3e5f768977a7825e70335feff2cfac56e54ac33b/lua/cmake-tools/terminal.lua#LL542C1-L560C6

But this will block neovim instance, we should find a non-blocking or async way to check this.

rohit-kumar-j:
Posible solution...?
If we do a background non-blocking `vim.wait()` that looks inside the build directory every n milliseconds until the currently running cmake produces the `CMakeCache.txt` or something like that which tells the status of the cmake process is completed, once it finds this (effectively a more verbose on_success() condition), maybe even the timestamp of the file... We can then chain the next command with a small delay. 

We will need to create new functions for checking status of files like how the codemodel targets are obatined...

Civitasv:
Okay, hopefully someone can anwser it. :smile:

rohit-kumar-j:
Seems like this can be accomplished by writing a [dummy query into a file](https://discourse.cmake.org/t/is-there-a-way-to-check-when-any-cmake-task-has-finished-purely-from-the-filesystem-build-directory/8271/4?u=j_rohit). 

@Civitasv, do you have any idea how we may do this?

Civitasv:
I've researched some plugins about terminal, cannot find a suitable solution to get notified when terminal executes commands successfully. But I find another way, See it at #63.

rohit-kumar-j:
Nice! That is a simple and effective solution :D

PR #63, opened 2023-06-10
This is a fix for #60

Civitasv:
This is a fix for #60, instead using `on_success` function to chain commands, now I use `&&` to join multiple commands. For example, If uses invoke `CMakeRun`, and not yet `CMakeGenerate` and `CMakeBuild`, it will turn like `cmake -B out/Debug -S . -D CMAKE_BUILD_TYPE:STRING=Debug -DCMAKE_EXPORT_COMPILE_COMMANDS=1 && cmake --build out/Debug --target TEST && c
d /home/civitasv/Documents/workflow/test/cmake-test/out/Debug && ./TEST `.

It actually works really good. But it still have a little flaw: For the very first time, users must call `CMakeGenerate` once to generate build system. Then it will work just like quickfix version. I think it's totally acceptable.

### DIFF
--- a/lua/cmake-tools/config.lua
+++ b/lua/cmake-tools/config.lua
@@ -110,7 +110,6 @@ function Config:check_launch_target()
   end
 
   -- 2. not select launch target yet
-  -- print("SELECTED", self.launch_target)
   if not self.launch_target then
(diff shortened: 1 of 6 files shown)


rohit-kumar-j:
Nice. This makes it a lot simpler and with no sync `vim.wait()`

PR #65, opened 2023-06-11
Quick* commands and code model info

hfn92:
I added some "quick" version of existing commands that allow cmake calls without having to change the current configuration.
`CMakeQuickBuild`,`CMakeQuickRun` and `CMakeQuickDebug`. They work by either passing the desired target or when none passed will query the user via `vim.ui.select`.  
`CMakeQuickBuild` is especially useful for cmake utility targets that produce no binary.

I also added `cmake.get_model_info` to query infos from the cmake model.  All well as `cmake.get_cmake_launch_targets`.

I'm writing a plugin for the use of gtest together with cmake and this allows to query some useful information about what executables exist, where they are and what cpp files they compile. As well as using `cmake-tools` to build specific tests and run them in debug mode.

I hope I didn't break anything (was quite tricky to merge after the recent changes) :pray:

### DIFF
--- a/lua/cmake-tools/config.lua
+++ b/lua/cmake-tools/config.lua
@@ -269,7 +269,10 @@ local function get_targets(config, opt)
       if (target_name_on_disk ~= nil) then -- only executables have name on disk?
         path = path .. "/" .. target_name_on_disk
       end
-      local abs_path = Config.build_directory .. "/" .. path
+      local abs_path = ""
+      if type == "executable" then
+        abs_path = Config.build_directory .. "/" .. target_info["artifacts"][1]["path"]
+      end
       if not (opt.only_executable and (type ~= "executable")) then
         if target_name == config.build_target then
           table.insert(targets, 1, target_name)


lua/cmake-tools/terminal.lua
```diff
@@ -427,7 +427,7 @@ function terminal.prepare_cmd_for_execute(executable, args, launch_path)
     -- Weird windows thing: executables that are not in path only work as ".\executable" and not "executable" on the cmdline (even if focus is in the same directory)
     full_cmd = full_cmd .. ".\\" .. executable
   elseif osys.islinux then
-    full_cmd = full_cmd .. "./" .. executable
+    full_cmd = " " .. full_cmd .. "./" .. executable -- adding a space in front of the command prevents bash from recording the command in the history (if configured)
```
hfn92, line 430: Oh I also added this. when `HISTCONTROL=ignoreboth` is set commands will not be saved in bash history.

Civitasv:
Very nice feature! And I am also very interested in your `cmake-gtest` plugin. I think it may be better to add this test feature to our `cmake-tools` plugin, but it's up to you, what do you think of that?

hfn92:
Thanks :+1:  I've edited the readme of `cmake-gtest`  to show what I'm going for. If you don't think this is out of scope for `cmake-tools` (since it's gtest specific) I wouldn't mind this being integrated into `cmake-tools`.  It would make development easier and I plan on adding `ctest` features as well

Civitasv:
Following discussion about test support will be at #68

PR #78, opened 2023-06-27
Added new  terminal filetype for easier filtering of terminal buffers 

rohit-kumar-j:
This is an option that can be provided for other plugins to integrate with cmake_tools. The scrolling buffer can be simply filtered by searching for 'cmake_tools_terminal' filetype directly. For example, using [`hardtime.nvim`](https://github.com/m4xshen/hardtime.nvim)

Ex:

```lua
-- Setter (done on our side)
vim.api.nvim_buf_set_option(<bufnr>, 'filetype', 'cmake_tools_terminal')

-- Getter (in another plugin. Example: hardtime.nvim)
local ft = vim.api.nvim_buf_get_option(<bufnr>, 'filetype')
```

PS: Can be helpful to eliminate the need to create prefixed named for all cmake terminals if we can filter with file types.


rohit-kumar-j:
New pr from another branch

PR #79, opened 2023-06-27
Added new terminal filetype for easier filtering of terminal buffers

rohit-kumar-j:
This is an option that can be provided for other plugins to integrate with cmake_tools. The scrolling buffer can be simply filtered by searching for 'cmake_tools_terminal' filetype directly. For example, using [`hardtime.nvim`](https://github.com/m4xshen/hardtime.nvim). You can specify the ignored/included filetypes for specific operations like: 

```lua
{
    "m4xshen/hardtime.nvim",
    event = "VeryLazy",
    opts = { disabled_filetypes = { "cmake_tools_terminal" },
    }
},

```

Ex:

```lua
-- Setter (done on our side)
vim.api.nvim_buf_set_option(<bufnr>, 'filetype', 'cmake_tools_terminal')

-- Getter (in another plugin. Example: hardtime.nvim)
local ft = vim.api.nvim_buf_get_option(<bufnr>, 'filetype')
```

PS: Can be helpful to eliminate the need to create prefixed named for all cmake terminals if we can filter with file types.

### DIFF
--- a/docs/howto.md
+++ b/docs/howto.md
@@ -469,6 +469,12 @@ By default, this plugin uses quickfix console for generate, build, clean, instal
 
 But if you want to always use terminal(for example, you want to record all commands and corresponding output), there is a way. You need set `cmake_always_use_terminal` to true, then, all commands will be executed in the terminal.
 
+---
+
+## Integrations
+
+### NvimTree
+
 For users that are using nvim-tree, to keep the size of terminal constant, you should add the following configuration for nvim-tree.
 
 ```lua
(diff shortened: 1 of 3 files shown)


rohit-kumar-j:
Tested: Works for both `cmake_always_use_terminals = true/false`

Civitasv:
I think there is no need to make an option. It’s okay to just give terminal this type. What do you think?

rohit-kumar-j:
I was thinking exactly that, but was on the fence with this. hehe..

Civitasv:
Make it default. 😄

rohit-kumar-j:
done

PR #82, opened 2023-07-10
Session init fixes

cjuub:
First of all, thanks for the session support! Was on my wish list for this plugin.

I got some errors running master tough, related to creation of the cache directory. On Linux the "touch" command fails if the cache directory does not already exist. So I added some code for that in this PR.

Also made sure that if session.load() returned empty table (first time starting new session), we don't overwrite the config with it as that broke stuff.

Commits in this PR resolved stuff for me, but I have not tested for any other OS.


Civitasv:
Thanks for your fix. :small_airplane:

PR #85, opened 2023-07-15
Add option to wrap cmake.run calls

hfn92:
I added an option to wrap `run` calls in additional arguments.

This allows for example to execute `perf record --call-graph dwarf {executeable}` or `valgrind {program}` with:

```lua
function RunPerf()
  local cmake = require "cmake-tools"
  cmake.run { wrap_call = { "perf", "record", "--call-graph", "dwarf" } }
end

function RunValgrind()
  local cmake = require "cmake-tools"
  cmake.run { wrap_call = { "valgrind", "--leak-check=full" } }
end

```

### DIFF
--- a/docs/howto.md
+++ b/docs/howto.md
@@ -463,6 +463,20 @@ it looks like:
 
 ![lualine UI](./images/2023-06-06-22-02-06.png)
 
+## Integration of external Tools
+
+`cmake-tools.run` can be used to run an executable with an external tool such as `valgrind` or `perf` by prepending the run call with custom arguments.
+
+This example creates a custom function that runs the currently selected launch target with `perf`
+
+```lua
+function RunPerf()
+  local cmake = require "cmake-tools"
+  cmake.run { wrap_call = { "perf", "record", "--call-graph", "dwarf" } }
+end
+```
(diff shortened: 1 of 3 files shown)


Civitasv:
Hi, can you explain why is it useful?

hfn92:
I think it's useful when having to often run analysis tools on your executables.  Since `cmake-tools` already knows how to run the targets (even with launch_args). I can just create simple wrapper functions that calls these tools with the requirement arguments instead of constructing them myself.
Eg `RunPerf` from above runs:
```
 cd /home/build_RelWithDebInfo/Tesst && perf record --call-graph dwarf ./Target launch_args
```

Civitasv:
Oh, that's useful. Can you also write a doc about it, maybe add a section in the [howto](https://github.com/Civitasv/cmake-tools.nvim/blob/master/docs/howto.md) file?

hfn92:
> Oh, that's useful. Can you also write a doc about it, maybe add a section in the [howto](https://github.com/Civitasv/cmake-tools.nvim/blob/master/docs/howto.md) file?

Do you think this is fine? [632a7ec](https://github.com/Civitasv/cmake-tools.nvim/pull/85/commits/632a7ec24d572173ba39c81f4a93e7d78bba123c)

Civitasv:
Yeah. LGTM.

PR #89, opened 2023-07-20
added cmake.get_build_directory

hfn92:
I added this to read coverage data from the build directory. Might also be useful for other data that gets written into build directory.


Civitasv:
LGTM

PR #92, opened 2023-07-23
Fixed start_insert and focus on terminal option

Ceraxel:
- fixed start insert option upon the launch of the terminal
- fixed inverted focus on terminal option

### DIFF
--- a/lua/cmake-tools/terminal.lua
+++ b/lua/cmake-tools/terminal.lua
@@ -169,7 +169,7 @@ function terminal.send_data_to_terminal(buffer_idx, cmd, opts)
     -- do nothing
   end
 
-  if opts and (opts.focus_on_launch_terminal or opts.focus_on_main_terminal) then
+  if opts and not (opts.focus_on_launch_terminal or opts.focus_on_main_terminal) then
     vim.cmd("wincmd p") -- Goes back to previous window: Equivalent to [[ CTRL-W w ]]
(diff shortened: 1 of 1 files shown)


lua/cmake-tools/terminal.lua
```diff
@@ -169,9 +169,9 @@ function terminal.send_data_to_terminal(buffer_idx, cmd, opts)
     -- do nothing
   end
 
-  if opts and (opts.focus_on_launch_terminal or opts.focus_on_main_terminal) then
+  if opts and not (opts.focus_on_launch_terminal or opts.focus_on_main_terminal) then
```
Civitasv, line 172: Why is it should be `not`?

Ceraxel: Originally, setting the option focus on terminal to true does the opposite behavior and does not actually focus the terminal upon launch while setting it to false does the intended behavior

Here is the snippet of the code I modified
`
  if opts and not (opts.focus_on_launch_terminal or opts.focus_on_main_terminal) then
    vim.cmd("wincmd p") -- Goes back to previous window: Equivalent to [[ CTRL-W w ]]
  elseif opts and opts.start_insert then
    vim.cmd("startinsert")
  end
`

Civitasv: okay, I see.

PR #94, opened 2023-07-24
fixed: terminal window focus issues

Ceraxel:
Fixed terminal focus issues wherein terminal window does not get focused if there is already an existing terminal window and when the start_insert option is disabled


Civitasv:
Thanks!

PR #96, opened 2023-07-27
Feature/add more executors

ErezAmihud:
Hi, this is a merge request that adds more options to run commands on, for example, [overseer](https://github.com/stevearc/overseer.nvim).
I may have went a bit overboard with the changes, in general I have made the terminal and quickfix to have an interface called `executor.Adapter` which contains the neccessery things to run commands.
I have made terminal and quickfix "classes"  (not sure about this choice) to avoid passing the cmake_opts to them in every place they are used.
In all the locations terminal and quickfix are used I have made sure to use the executor.

This is not fully tested, I intend to use it in my setup so If there are things that break I would fix them, I am sure somethings broke by this change, and it probably should be properly tested, and community tested before moving to master (if ever).


Civitasv:
Thanks for your contributing, that's a really nice idea which can unify execution in Neovim, using this, we can do more than just cmake generation. I will review it as soon as I can, it is a big change after all.

Civitasv:
Hi, I've merged it with #101, something may broke, can you both see if it works? Thanks! @ErezAmihud @hfn92

hfn92:
@Civitasv Sorry not sure how to test it :/ I don't know how to checkout your changes. `gh pr checkout 96` only seems to check out the original branch

Civitasv:
> but I prefer to use pure function rather than class for executors, it is not that intuitive. 

> Oh looks like you removed some of your commmits

@hfn92 Yes, I find that it is not very intuitive using class, it's better to just use pure function instead in my opinion.

Civitasv:
@ErezAmihud And the design idea here is: for `CMakeRun`, we always use terminal, for others, we can use terminal, quickfix or other "executors" like what you have done with overseer.

Civitasv:
> to avoid passing the cmake_opts to them in every place they are used.

And one of the reason about it is we want to keep exactly one instance of const and config, so if we can handle per-instance properly, we don’t need passing conf every place.

ErezAmihud:
> > but I prefer to use pure function rather than class for executors, it is not that intuitive.
> 
> > Oh looks like you removed some of your commmits
> 
> @hfn92 Yes, I find that it is not very intuitive using class, it's better to just use pure function instead in my opinion.

I agree that pure functions are better here, but I don't agree with passing multiple arguments to every function call, it would be hard to accomedate new executors this way.

What do you thing about having an `executor` member of `config` that contains a table with the name of the executor and executor specific parameters, and then have a util function to get the corresponding executor and call it's functions with the config.executor table?

```lua
-- open executor window
executor = util.get_executor(config.executor.name)
executor.open(config.executor)

---@param name string
function util.get_executor(name)
    return require("cmake-tools.executors")[name]
end
```

This will make sure that no executor specific parameters are not used in the `init` file, and it would give us pure functions in the executor modules.

Civitasv:
That seems a cleaner way.

Civitasv:
It is
> ```lua
> -- open executor window
> executor = util.get_executor(config.executor.name)
> executor.open(config.executor.opts) — changes here
> 
> ---@param name string
> function util.get_executor(name)
>     return require("cmake-tools.executors")[name]
> end
> ```

right?

ErezAmihud:
> It is
> 
> > ```lua
> > -- open executor window
> > executor = util.get_executor(config.executor.name)
> > executor.open(config.executor.opts) — changes here
> > 
> > ---@param name string
> > function util.get_executor(name)
> >     return require("cmake-tools.executors")[name]
> > end
> > ```
> 
> right?

Yes

Civitasv:
And besides the interface for `Executor`, I think it's also clearer to also have interface for `Runner`.

**Executor:** Used for cmake generate, build, clean, install, stop, etc
**Runner:** Used for cmake run, debug

This way, `terminal` can be both `Executor` and `Runner`, `quickfix` mainly is a `Executor`, though can be also used as a `Runner`(have issue when need input from stdin).

ErezAmihud:
> And besides the interface for `Executor`, I think it's also clearer to also have interface for `Runner`.
> 
> **Executor:** Used for cmake generate, build, clean, install, stop, etc **Runner:** Used for cmake run, debug
> 
> This way, `terminal` can be both `Executor` and `Runner`, `quickfix` mainly is a `Executor`, though can be also used as a `Runner`(have issue when need input from stdin).
> 
> Anyway, Take your time ~

It would be nice to have it, but I think it would be out of scope for this cr.

> And besides the interface for `Executor`, I think it's also clearer to also have interface for `Runner`.
> 
> **Executor:** Used for cmake generate, build, clean, install, stop, etc **Runner:** Used for cmake run, debug
> 
> This way, `terminal` can be both `Executor` and `Runner`, `quickfix` mainly is a `Executor`, though can be also used as a `Runner`(have issue when need input from stdin).
> 
> Anyway, Take your time ~

It makes sense, but I think it may be out of scope for this cr.
Same as creating a proper way to get the command line to run - instead of using the full_command global variable.

Anyway, I updated the pr, 2 things are left: close an open terminal, adding the notifications, updating the docs and figure out how we want the config to look (as the setup currently is it combines the executor.opts with the new ones).

lua/cmake-tools/config.lua
```diff
@@ -18,6 +18,8 @@ local Config = {
   kit = nil,
   configure_preset = nil,
   build_preset = nil,
+  executor = nil,
+  terminal = nil,
```
Civitasv, line 22: There should be executor and runner in the future. Now it is a little confusing.

Civitasv: But for this PR, it's okay.

lua/cmake-tools/const.lua
```diff
@@ -20,28 +20,33 @@ local const = {
...
-    start_insert_in_launch_task = false, -- If you want to enter terminal with :startinsert upon using :CMakeRun
-    start_insert_in_other_tasks = false, -- If you want to enter terminal with :startinsert upon launching all other cmake tasks in the terminal. Generally set as false
-    focus_on_main_terminal = false, -- Focus on cmake terminal when cmake task is launched. Only used if cmake_always_use_terminal is true.
-    focus_on_launch_terminal = false, -- Focus on cmake launch terminal when executable target in launched.
+      -- Running Tasks
+      start_insert_in_launch_task = false, -- If you want to enter terminal with :startinsert upon using :CMakeRun
+      start_insert_in_other_tasks = false, -- If you want to enter terminal with :startinsert upon launching all other cmake tasks in the terminal. Generally set as false
+      focus_on_main_terminal = false, -- Focus on cmake terminal when cmake task is launched. Only used if cmake_always_use_terminal is true.
+      focus_on_launch_terminal = false, -- Focus on cmake launch terminal when executable target in launched.
+    },
```
Civitasv, line 49: I like this way to organize options for executor and terminal.

lua/cmake-tools/init.lua
```diff
@@ -20,1075 +19,1018 @@ local full_cmd = ""
...
-      config.launch_target = old_config.launch_target
-      config.launch_args = old_config.launch_args
-      config.kit = old_config.kit
-      config.configure_preset = old_config.configure_preset
-      config.build_preset = old_config.build_preset
-    end
-  end
+	if values.cmake_executor ~= nil then
+		const.cmake_executor = values.cmake_executor
+	end
```
Civitasv, line 24: Why should we handle `cmake_executor` specifically?

Civitasv:
Please firstly rebase master, cause there are many changes since you made this PR. Or it maybe painful to merge.

lua/cmake-tools/init.lua
ErezAmihud: Changed, but it still does have a specific way to handle it, because using deepcopy with it will mix options from multiple executors

ErezAmihud:
> Please firstly rebase master, cause there are many changes since you made this PR. Or it maybe painful to merge.

Done, it is up to date with current master

lua/cmake-tools/executors/overseer.lua
```diff
@@ -0,0 +1,59 @@
+local overseer = require("overseer")
```
Civitasv, line 1: You should use `pcall` to check if overseer is installed.

lua/cmake-tools/quickfix.lua
```diff
@@ -99,44 +59,36 @@ function quickfix.run(cmd, env, args, opts)
...
-        icon = ""
-        if opts.on_success then
-          opts.on_success()
-        end
-      elseif opts.cmake_quickfix_opts.show == "only_on_error" then
-        quickfix.show(opts.cmake_quickfix_opts)
+      if code == 0 and opts.show == "only_on_error" then
+        quickfix.show(opts)
         quickfix.scroll_to_bottom()
       end
```
Civitasv, line 78: Compare to previous, I think this logic has problem. If opts.show equals only_an_error, we should only show quickfix window when code not equals zero

lua/cmake-tools/init.lua
```diff
@@ -38,6 +38,18 @@ function cmake.setup(values)
     telescope.load_extension("cmake_tools")
   end
   const = vim.tbl_deep_extend("force", const, values)
+  if const.cmake_executor.name == "terminal" then
+    const.cmake_executor = const.cmake_terminal
+    if const.cmake_notifications.enabled then
+      log.info("Cannot use cmake-tools notifications in terminal mode")
```
Civitasv, line 44: Cannot use cmake-tools notifications in terminal mode -> Currently we don't support using cmake-tools notifications in terminal mode

lua/cmake-tools/init.lua
```diff
@@ -122,9 +134,9 @@ function cmake.generate(opt, callback)
     vim.list_extend(args, config.generate_options)
     vim.list_extend(args, fargs)
 
-    local env = environment.get_build_environment(config, const.cmake_always_use_terminal)
+    local env = environment.get_build_environment(config, config.executor.name == "terminal")
```
Civitasv, line 137: Maybe we should save this flag `config.executor.name == "terminal"` to config.

Civitasv:
This looks pretty good now, just several little problems.

ErezAmihud:
@Civitasv I changed all the comments :)
For error handling I added an executor.is_installed function which returns an error string or nil, for the `setup` function to use.

Civitasv:
Thanks for your contribution!

Issue #90, opened 2023-07-20
Configure target specific environment variables


hfn92:
**Describe the problem or limitation you are having**
I sometimes need to run executables with specific environment variables. 
This could also be useful for generating and building as some compilers (qnx?) require specific env-vars to compile.

**Describe the solution you'd like**
Inspired by how QtCreator handles environment variables I think it would be useful to be able to configure env-vars for the generate/configure/build step and for individual targets. 

My idea is that the env-vars could be configured in a regular vim buffer using lua code. (Similar to the quick window of [harpoon](https://github.com/ThePrimeagen/harpoon))

For example. `CMakeConfigureBuildEnvironment` would open up a floating window with:
```lua
return {
  env { 
    VERBOSE = 1,
  }
}
```
This would add `VERBOSE=1` to env-vars for the generate and build steps

And `CMakeConfigureRunEnvironment {target_or_current}` for target specific env-vars.
```lua
return {
  inherit_build_environment = true,
  env { 
    log_file = "/tmp/log.txt",
    log_level = "trace",
  }
}
```
This would add `VERBOSE=1 log_file = "/tmp/log.txt" log_level = "trace"`  to env-vars when running the specific target.

This would also be nice since `cmake-tools` could persist these similar to session.

**Describe alternatives you've considered**
I can just close nvim set env-vars.

Civitasv:
That's an interesting idea! Currently, for generate step, this plugin already supports setting env by using [kits](https://github.com/microsoft/vscode-cmake-tools/blob/main/docs/kits.md#general-options) or [variants](https://github.com/microsoft/vscode-cmake-tools/blob/main/docs/variants.md#variant-options)(though we don't support env option yet). So we should consider the precedence of them.

For build and run step, currently there is no convenient way to do this. Maybe we can try to implement your idea.

hfn92:
Since since you seem to be okay with the general idea of the feature I started with a bit of a Prototype.

https://github.com/hfn92/cmake-tools.nvim/commits/env-feature

This should already work for launch target but not for building and debugging. And probably only for unix but its a start

Civitasv:
Bravo 👏

rohit-kumar-j:
@hfn92 , this is a good feature, and I would like to test it out fully for windows. Do you have a test project covering all the test cases for this?

[...]

PR #97, opened 2023-07-27
Configure target specific environment variables

hfn92:
I implemented something like discussed here https://github.com/Civitasv/cmake-tools.nvim/issues/90

Could you try that out and tell me what you think?

I also marked this as draft since the terminal env vars won't work for windows  and I currently can't test that.  (Maybe this could work? https://superuser.com/a/1405733 - running commands in a sub shell with `cmd /c ` or just live with the fact that env war won't be reset?


hfn92:
I implemented something for window using `cmd /c`. It worked for me. But I'm not sure if it would break something

hfn92:
forgot to test debugging ... will open pr again later

PR #101, opened 2023-07-28
nvim-notify support

hfn92:
I added support for [nvim-notify](https://github.com/rcarriga/nvim-notify).
When nvim-notify is installed this add some eyecandy when executing cmake commands that use the quickfix window.
Should pair well with `cmake_quickfix_opts.show  =  "only_on_error"`

[cmake-notify.webm](https://github.com/Civitasv/cmake-tools.nvim/assets/26342294/414a01f1-1b64-4ce1-aee3-945e743c5726)


Civitasv:
Wow, it's really cool!

PR #102, opened 2023-07-28
run stylua on all files

hfn92:
since there is a github action now that checks if all files in a PR are formatted the master should be already be formatted


Civitasv:
I thought it will automatically format, isn’t i?

PR #103, opened 2023-07-29
Fix user preset detection for kebab cased files

hrzlgnm:
Tested locally by renaming my CMakePresets.json and CMakeUserPresets.json to their kebab-case counter parts, my user preset was still found.  But there was one issue. Upon invocation of CMakeRun, cmake itself complained that it could not find such a preset.

### DIFF
--- a/lua/cmake-tools/presets.lua
+++ b/lua/cmake-tools/presets.lua
@@ -14,10 +14,12 @@ function presets.check()
     local file = nil
     local presetFiles = {}
     for _, f in ipairs(files) do -- iterate over files in current directory
-      if f == "CMakePresets.json" or
-          f == "CMakeUserPresets.json" or
-          f == "cmake-presets.json" or
-          f == "cmake-user-presets.json" then -- if a preset file is found
+      if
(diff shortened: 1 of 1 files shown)


Civitasv:
> Upon invocation of CMakeRun, cmake itself complained that it could not find such a preset.

Di you find why that happens?

hrzlgnm:
> > Upon invocation of CMakeRun, cmake itself complained that it could not find such a preset.
> 
> Di you find why that happens?

I was dumb, i didn't use CMakeGenerate beforehand, now I did and it worked well :)

Civitasv:
> I was dumb, i didn't use CMakeGenerate beforehand, now I did and it worked well :)

Do you set cmake_always_use_terminal to true?

hrzlgnm:
> > I was dumb, i didn't use CMakeGenerate beforehand, now I did and it worked well :)
> 
> Do you set cmake_always_use_terminal to true?

yep i have it enabled

Civitasv:
Okay, I understand.

Issue #100, opened 2023-07-28
Allow working directory to specified when running executables


jla2000:
**Describe the problem or limitation you are having**
I need to set the working directory of some cmake targets to something else. I'm not sure on how to do this right now as it will always choose the build folder where the executable is run with CMakeRun as the working directory.

**Describe the solution you'd like**
Something similiar to CMakeLaunchArgs, like CMakeLaunchDir where an absolute or relative (eg ../) path to the build folder can specified for both simply running the executable or debugging with nvim-dap.

**Describe alternatives you've considered**
Right now I have no real alternative but to manually run with the adopted working dir, but eg for debugging it would be really nice to have it just working out of the box.

**Additional context**
--

jla2000:
> You can change current working dir by change “vim.loop.cwd()”.

Just override this function and return a string pointing to the working directory?

Civitasv:
An example: https://vi.stackexchange.com/questions/34871/set-current-working-directory-when-opening-vim.

hfn92:
Doesn't `cmake-tool` always `cd` into the directory of the executable before running? I think OP is asking about allowing somethin like
`cd /custom/dir && path/to/exe/exe`
instead of
`cd path/to/exe/ && exe`

Civitasv:
@hfn92 You are right, I misunderstood.

jla2000:
@hfn92 Yes, this is what I meant. I would like to be able to launch the executable with another cwd, not directly the path where the executable is stored. It would be nice if this could be also like a relative path so you could pass ../ for example to reference the parent folder of the binary

jla2000:
So what would could work is cd /path/to/exe && cd custom/dir && path/to/exe/exe

Then it would be possible to specify a relative or absolute path with custom/dir

hfn92:
I wonder if a nice way to configure this would be using the same way I intend to with with this feature https://github.com/Civitasv/cmake-tools.nvim/issues/90
This could also be used to configure launch arguments
```lua
return {
  working_drectory = "../",
  args = [[arg1 arg2]],
  inherit_build_environment = true,
  env { 
    log_file = "/tmp/log.txt",
    log_level = "trace",
  }
}
```

I think this would be a really convenient way to configure these things, since it works directly in a vim buffer and is easy to persist

Civitasv:
> I think this would be a really convenient way to configure these things, since it works directly in a vim buffer and is easy to persist

That’s indeed a convenient way!

[...]

PR #104, opened 2023-07-29
Configure target specific environment variables

hfn92:
As discussed here: https://github.com/Civitasv/cmake-tools.nvim/issues/90

Works for me but probably requires some additional testing. Especially for windows users.


Civitasv:
I will test it soon.

hfn92:
> I will test it soon.

Thanks :) Sorry for the PR spam :pray: 

I was wondering. Before you merge it, should I adapt it for what was discussed here? https://github.com/Civitasv/cmake-tools.nvim/issues/100 Right know the config window is only intended for env vars?

Civitasv:
> Thanks :) Sorry for the PR spam

Don’t be sorry, thanks for your contribution to make it better. 😄 

> Before you merge it, should I adapt it for what was discussed here?

I think it’s a good idea to implement now, then I can review all of them.

hfn92:
> I think it’s a good idea to implement now, then I can review all of them.

Ok will do. This is not a major change.  Just some renaming. So you are still free to review the current state :)

Civitasv:
Okay 👍

Civitasv (approved):

hfn92:
I just noticed that its probably better to persist settings as table instead of a string ...

so instead of
```lua
base_settings = "return {\n  env = {\n    VERBOSE=1\n  }\n}\n",
```
it should be like this (in session config)
```
base_settings = {  env =  { VERBOSE=1 } },
```

What do you think? I could still change that

Civitasv:
I agree with you, table is more primitive in lua.

hfn92:
> I agree with you, table is more primitive in lua.

A disadvantage is that comments won't be saved. A user might want to uncomment something for later use .

Civitasv:
That's a case. Suddenly I think the string format maybe better lol. What do you think?

hfn92:
Its tricky huh :D 
A table would have the advantage that it can be easily extended with missing settings if we add new ones. 
We could use strings in memory but persist tables. So comments would exist for the current instance but not be saved.
I'll think about it a bit

Civitasv:
> We could use strings in memory but persist tables. So comments would exist for the current instance but not be saved.

yeah, it’s better.

hfn92:
I decided against using both. Its easier to manage this way and more consistent between usage and allows for easier detection if something is wrong. And we can extend the table with defaults if something is missing at runtime.  But switching to using strings is always possible later.

Civitasv:
Yes, I agree with you.

Civitasv:
Please fix the conflicts then we can merge to master.

hfn92:
> Please fix the conflicts then we can merge to master.

I tried twice locally. Not sure why it didn't work :/ Worked on gh directly though

Civitasv:
And you should fix the format

PR #106, opened 2023-07-29
Add Telescope extension

hfn92:
I added an extension for [telescope.nvim](https://github.com/nvim-telescope/telescope.nvim).

from docs:

`Telescope cmake_tools` shows files associated with the cmake project. Ignoring some files such as objects files or cmake rules. (Combines both `sources` and `cmake_files`).

`Telescope cmake_tools sources` shows only source files and files directly added to targets.

`Telescope cmake_tools cmake_files` shows files associated with the cmake-model (CMakeLists files and similar).

Additionaly `CMakeShowTargetFiles` can be used to only show files associated with a specific target.

I mainly added this because I sometimes need to look at file that are outside of the source tree. Either generated files or files or files included via a package mananger. `Telescope find_files` doesn't know about these files but the cmake code model does.


lua/cmake-tools/config.lua
```diff
@@ -61,9 +61,37 @@ function Config:generate_query_files()
...
+        nil,
+        "Unable to create file " .. cmakeFiles_file.filename
+      )
+    end
+  end
+
   return Result:new(Types.SUCCESS, true, "yeah, that could be")
 end
 
+function Config:get_cmakeFiles()
```
Civitasv, line 79: Why not `get_cmake_files`

hfn92: whoops

Issue #109, opened 2023-07-29
Telescope cmake_tools throws


Amleto:
**Bug description**

I'm not sure if there are some undocumented pre-reqs or anything?

Run `:Telescope cmake_tools` gives:
```
Error executing Lua callback: ...ta/lazy/cmake-tools.nvim/lua/cmake-tools/file_picker.lua:18: bad argument #1 to 'ipairs
' (table expected, got nil)
stack traceback:
        [C]: in function 'ipairs'
        ...ta/lazy/cmake-tools.nvim/lua/cmake-tools/file_picker.lua:18: in function 'get_files_from_target'
        ...ta/lazy/cmake-tools.nvim/lua/cmake-tools/file_picker.lua:51: in function 'get_source_files'
        ...ake-tools.nvim/lua/telescope/_extensions/cmake_tools.lua:30: in function 'fn'
        ...ake-tools.nvim/lua/telescope/_extensions/cmake_tools.lua:16: in function <...ake-tools.nvim/lua/telescope/_ex
tensions/cmake_tools.lua:8>
        .../nvim-data/lazy/telescope.nvim/lua/telescope/command.lua:193: in function 'run_command'
        .../nvim-data/lazy/telescope.nvim/lua/telescope/command.lua:253: in function 'load_command'
        ...Local/nvim-data/lazy/telescope.nvim/plugin/telescope.lua:108: in function <...Local/nvim-data/lazy/telescope.
nvim/plugin/telescope.lua:107>
```

**Steps to reproduce**
I have telescope and cmake_tools plugins loaded. My current dir is a small c++/cmake repo. I run.
`:Telescope cmake_tools`

<details><summary>Minimal configuration</summary>
</details>

**Expected behavior**
The telescope command executed ok. 

**Screenshots**

**Environment**

- OS: Win10
- Plugins commit hash:  8e8e9520bf09c52f540b8c12ae29

Civitasv:
@hfn92 can you see it

hfn92:
can you upload the contens of `./cmake/api` ? the whole project would be even better of course

hfn92:
Still I think this should fix it https://github.com/Civitasv/cmake-tools.nvim/pull/110
But would be good to know how the cmake api response looks

Amleto:
> can you upload the contens of `./cmake/api` ? the whole project would be even better of course

I'm sorry, I don't know what that refers to.

[...]

PR #110, opened 2023-07-29
Fix telescope extension throw

hfn92:
See https://github.com/Civitasv/cmake-tools.nvim/issues/109


Civitasv:
Thanks for the quick fix.

PR #111, opened 2023-07-29
save and configure launch_args in target_setting

hfn92:
`CMakeLaunchArgs` still works but now it can also be configured with the config window


Civitasv:
1. You forgot to add CMakeSettings and CMakeTargetSettings to the docs
2. Considering we already have CMakeTargetSettings, is it necessary to also have CMakeLaunchargs?

Civitasv:
> 2\. Considering we already have CMakeTargetSettings, is it necessary to also have CMakeLaunchargs?

Forget about it, it still is useful to have both.

hfn92:
> 1. You forgot to add CMakeSettings and CMakeTargetSettings to the docs

Done. Also made a new page for this.

Issue #108, opened 2023-07-29
Improve docs to help with compiler output parsing in quickfix list


Amleto:
**Describe the problem or limitation you are having**
error messages from compiler output are not parsed in quickfix list (windows, msbuild)

**Describe the solution you'd like**
Some 'windows users please note:' section in the readme/docs

**The required action**
```lua
-- MSBuild:
vim.opt.errorformat:append([[\ %#%f(%l\,%c):\ %m]])
-- cl.exe:
vim.opt.errorformat:append([[\ %#%f(%l)\ :\ %#%t%[A-z]%#\ %m]])
```

Civitasv:
Thanks for your suggestion. What does it look now?

Amleto:
Now (default):
![image](https://github.com/Civitasv/cmake-tools.nvim/assets/2996796/c75549c2-3da5-4aa9-bb44-e8fa44e92d38)
where quickfix navigation (`:cnext`. `:cprevious`) _do not_ work.

With my extra errorformat:
![image](https://github.com/Civitasv/cmake-tools.nvim/assets/2996796/91426e53-bdbe-4db3-8bb1-6ebb669a13b1)
where quickfix navigation _does_ work.

Civitasv:
![image](https://github.com/Civitasv/cmake-tools.nvim/assets/37768049/feb2b091-2666-4655-afe5-ae502ebc5e51)
By default, Linux looks like this.

Civitasv:
Can you make a PR to describe your solution? I think it's suitable to add this to the [howto](https://github.com/Civitasv/cmake-tools.nvim/blob/master/docs/howto.md) docs.

PR #116, opened 2023-07-30
#108 Update how-to: Add MSBuild errorformat to fix quickfix

Amleto:
#108 Update how-to: Add MSBuild errorformat to fix quickfix


Civitasv:
Thanks!

PR #117, opened 2023-07-30
Allow working directory to specified when running executables

hfn92:
See https://github.com/Civitasv/cmake-tools.nvim/issues/100

It now always runs executeables with absolute path. But this should be ok.


Civitasv:
Thanks’

Issue #120, opened 2023-07-31
Not all presets showing in select menu


Amleto:
**Bug description**

CMakeSelectConfigurePreset is not giving all the available options. Available from cli:
![image](https://github.com/Civitasv/cmake-tools.nvim/assets/2996796/f10a3f7e-d1d2-4940-b9eb-42f8770694f2)

cmake-tools:
![image](https://github.com/Civitasv/cmake-tools.nvim/assets/2996796/fdcc2eb8-1bcd-4b80-a8f7-b3976dea7da3)

**Steps to reproduce**
Unsure exactly on the minimum steps. Running into this issue at work. Seems like if there is a CMakeUserPreset then only that is used - everything from CMakePresets is ignored.

**Expected behavior**
All the presets from CMakePresets and CMakeUserPresets should be available to select.

**Environment**

- OS: win10
- Shell git-bash

PR #121, opened 2023-07-31
fix preset decode logic to not skip implicit include

Amleto:
#120 Fix parsing implicit include presets.


Civitasv:
Thanks!

hrzlgnm:
> #120 Fix parsing implicit include presets.

nice catch!
sorry i messed that up.

PR #126, opened 2023-08-06
Feature/change cwd

ErezAmihud:
Add an option to change the cwd used in build and run.
Currently this adds support to presets, but not to kits, variants as I don't know how to use them and have no minimal working model of their usage.

I can add it to kits/variants if someone have any minimal example, or someone else can do it in the future.


Civitasv:
I wonder why not just use `:cd <path>` instead?

Civitasv:
Okay, I understand, will review soon.

Civitasv:
Changes LGTM. 

> I can add it to kits/variants if someone have any minimal example, or someone else can do it in the future.

Kits and variants are definitely at root of cwd, just like presets. They are parsed using `kits.parse` and `variants.parse()`.

ErezAmihud:
> Changes LGTM.
> 
> > I can add it to kits/variants if someone have any minimal example, or someone else can do it in the future.
> 
> Kits and variants are definitely at root of cwd, just like presets. They are parsed using `kits.parse` and `variants.parse()`.
> 
> Edit: for kits, we provide a `cmake_kits_path` option to specify global kits path, if local kits don't exists, we use it, else we use local kits.

Ok, added support for both variants and kits, with the new cwd

lua/cmake-tools/init.lua
```diff
@@ -1238,7 +1294,7 @@ function cmake.compile_commands_from_soft_link()
     return
   end
 
-  local source = vim.loop.cwd()
+  local source = config.working_dir
     .. "/"
     .. config.build_directory.filename
     .. "/compile_commands.json"
```
Civitasv, line 1300: Why should destination be `vim.loop.cwd() .. "/compile_commands.json"` instead of `config.working_dir .. "/compile_commands.json"`

Civitasv:
And I've pushed a fix for macro expansion for presets: b490713bc661613a14f01d2d01a1c44382e1c5c3.

lua/cmake-tools/init.lua
Civitasv: Make sense

PR #128, opened 2023-08-08
Remove goto statements to ensure compatibility with (plain) Lua 5.1

pearzt:
Neovim is generally used with LuaJIT which [offers a set of extensions over Lua 5.1](http://luajit.org/extensions.html), including `goto`. However, Neovim can also be [built using plain Lua 5.1](https://github.com/neovim/neovim/blob/0211f889b9538f7df5fbcb06273d273fb071efff/cmake.deps/deps.txt#L10C1-L10C49). To allow the use of Neovim and cmake-tools.nvim on ISA's which are not supported by LuaJIT, it would be useful to keep the code compatible to plain Lua.

This PR removes all usages of `goto` and replaces them with simple conditionals.

### DIFF
--- a/lua/cmake-tools/environment.lua
+++ b/lua/cmake-tools/environment.lua
@@ -36,9 +36,7 @@ local function unroll(env, escape)
       table.insert(res, var)
     else
       -- unsupported type
-      goto ignore
     end
-    ::ignore::
   end
 
   return res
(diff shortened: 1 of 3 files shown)


Civitasv:
Make sense, please fix the format error.

pearzt:
Done

Civitasv:
Thanks!

Issue #129, opened 2023-08-11
Issues with :CMakeSelectCwd / finding preset files


jla2000:
**Bug description**
When updating to the latest version with this newly added `:CMakeSelectCwd` feature, I'm no longer able to generate or run anything. The plugin fails to do anything with this error message:
```
Error executing Lua callback: Vim:E484: Can't open file /home/user/code/microsar-adaptive//home/user/code/microsar-adaptive/CMakePresets.json
stack traceback:
	[C]: in function 'readfile'
	...e/nvim/lazy/cmake-tools.nvim/lua/cmake-tools/presets.lua:76: in function 'decode'
	...e/nvim/lazy/cmake-tools.nvim/lua/cmake-tools/presets.lua:105: in function 'parse'
	...hare/nvim/lazy/cmake-tools.nvim/lua/cmake-tools/init.lua:927: in function <...hare/nvim/lazy/cmake-tools.nvim/lua/cmake-tools/init.lua:913>
```

Notice that the mentioned path to the CMakePresets.json is not correct anymore here, looks like some broken concatenation.
Deletion of the session file does also not help.

A workaround I found to currently circumvent this is set `:CMakeSelectCwd` to `./`, then it works just fine. (Before it was `/home/user/code/microsar-adaptive`)

Civitasv:
@ErezAmihud Can you see it?

jla2000:
I have found the issue i think

ErezAmihud:
> I have found the issue i think

Can you send your presets file?

jla2000:
@ErezAmihud 
> Can you send your presets file?

I'm don't think so, its not public Xd

Have a look on my pr #130 I think it was an issue there with the path construction of these preset files.

jla2000:
@ErezAmihud 
Basically there was an absolute path constructed and checked and then added to the list of other relative files.
So when constructing the final filename the added files get appended the parent path again

jla2000:
Its merged now, so I'm closing this now

ErezAmihud:
> @ErezAmihud Basically there was an absolute path constructed and checked and then added to the list of other relative paths. So when constructing the final filename the added files get appended the parent path again

Yes, I looked at the pr now, good solution :)

PR #130, opened 2023-08-11
Fixed preset.decode creating wrong preset paths

jla2000:
The path will be concatenated twice as the includes will already contain the parent path.
(See #129)


jla2000:
I'm not sure about this paths included directly from the loaded file tho (line 58), these should be probably still loaded with an absolute path

jla2000:
Now its better, files are always included in the list only with their filename without an absolute path and then the final path is just constructed on read

Civitasv:
Thanks for your fix.

PR #132, opened 2023-08-18
Fix compatibility with windows path

ErezAmihud:
the paths on windows are `\` yet the  `get_path` function used the `/` seperator.


Civitasv:
Thanks!

PR #135, opened 2023-08-21
Fix broken string.find in clangd_on_new_config

cjuub:
Comparison was done without escaping the dash
characters. This resulted in the argument
--compile-commands-dir being added to clangd
command multiple times.

### DIFF
--- a/lua/cmake-tools/init.lua
+++ b/lua/cmake-tools/init.lua
@@ -1282,7 +1282,7 @@ function cmake.clangd_on_new_config(new_config)
   local found = false
   local arg = "--compile-commands-dir=" .. config.build_directory.filename
   for _, v in ipairs(new_config.cmd) do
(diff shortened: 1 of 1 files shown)


Civitasv:
I don't quite understand this error, can you explain more about it?

cjuub:
Sure, `string.find` uses pattern matching, and `-` is a special character when matching and must be escaped since we want to compare against actual `-` in the string.

The comparison will always be false. Even after --compile-commands-dir has been added by previous call to the hook.

The problem for me was that --compile-commands-dir argument got added to my clangd command each time a new buffer attached to the lsp. This can be seen in the output of "LspInfo" after opening several source files that clangd attached to.

Civitasv:
Okay, I understand.

For future reference: https://www.lua.org/pil/20.2.html.

>  -: also 0 or more repetitions

PR #136, opened 2023-08-25
Setting cwd with command args

LoKolbasz:
The CMakeSelectCwd command now accepts an optional argument which, if given, will set the CWD to the path of the arg. If no argument is given, the previous behavior applies.


Civitasv:
Thanks for your contribution!

PR #140, opened 2023-08-26
Fix broken build directory in session

cjuub:
When a session is loaded, it is assumed that
the build directory path is relative to the
project/cwd. However, when the session is saved,
then the build directory is serialized as an
absolute path. This caused the resulting path to
be "project root" + "full path to build directory":

/path/to/project//path/to/project/build/

This fixes the issue by serializing the build
directory as relative the project root.


Civitasv:
Thanks!

PR #147, opened 2023-09-02
Fix the clangd update to remove the old arg and add the new one

Elsa002:
When i used this plugin and changed the build type, it only took affect after restarting vim (restarting clangd did not help)

After looking at the code i found it only adds the compile commands path once, and after that it never updates it.

This pull request makes it so every time we update the build type/preset, it will remove any flag for a compile commands, and add the updated one, to ensure it is the only one passed to clangd.

This allows me to change from release to debug without restarting neovim, and seeing the changes happen immediately.


Civitasv:
Thanks for your contribution!

Issue #148, opened 2023-09-05
Changing buildPreset does not update selected configurePreset


lceWolf:
**Bug description**
Selecting a buildPreset with an associated configurePreset does not update the selected configurePreset

**Steps to reproduce**

<details><summary>Minimal configuration</summary>

```
{
  "version": 3,
  "cmakeMinimumRequired": {
    "major": 3,
    "minor": 19,
    "patch": 0
  },
  "configurePresets": [
    {
      "name": "foo",
      "cacheVariables": {
        "FOO": true
      }
    },
    {
      "name": "bar",
      "cacheVariables": {
        "BAR": false
      }
    }
  ],
  "buildPresets": [
    {
      "name": "build-foo",
      "configurePreset": "foo"
    },
    {
      "name": "build-bar",
      "configurePreset": "bar"
    }
  ]
}
```

</details>

1. Select configurePreset `foo` and buildPreset `build-foo` (:CMakeSelectConfigurePreset and :CMakeSelectBuildPreset)
2. Change buildPreset to `build-bar` (:CMakeSelectBuildPreset)

**Expected behavior**
The selected configurePreset is changed accordingly, but it stays the same

Issue #125, opened 2023-08-04
Check cache files if build directory is removed


deeedob:
Just encountered this: cmake-tools thinks it's in an active environment when a previously active build directory gets suddenly removed.

There was no way for me force it to re-generate. Even with `CMakeGenerate!`. So I had to remove the cache file manually. Perhaps this can be checked?

I used a CMakePresets.json file here and changed the name. Then just deleted the whole build dir.

- OS: Arch Linux
- Plugins commit hash: 20082f85909864a3c93acbc51c20dddc1004f39d

Civitasv:
Can you provide a detailed way to reproduce this problem?

Civitasv:
The latest commit should fix this problem. You should use `CMakeGenerate!` to re-generate.

Civitasv:
Closing now, feel free to open it if you encounter this error again.

PR #149, opened 2023-09-06
update configurePreset when changing buildPreset

lceWolf:
A buildPreset can have a configurePreset associated with it. When changing the buildPreset, the configurePreset should be updated accordingly.

After updating the configurePreset, we have to re-generate using the bang option as the new configurePreset might differ in the set cacheVariables and we have to get rid of the cached variables of the previous preset.

Question: should `select_configure_preset` use the bang option as per default as well? Keeping the `cacheVariables` values from the previous preset cached was unexpected to me 😅 

Fixes https://github.com/Civitasv/cmake-tools.nvim/issues/148#issue-1881394109

### DIFF
--- a/lua/cmake-tools/init.lua
+++ b/lua/cmake-tools/init.lua
@@ -968,8 +968,21 @@ function cmake.select_build_preset(callback)
         if config.build_preset ~= choice then
           config.build_preset = choice
         end
+        local associated_configure_preset =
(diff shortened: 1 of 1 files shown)


Civitasv:
> Question: should `select_configure_preset` use the bang option as per default as well? Keeping the `cacheVariables` values from the previous preset cached was unexpected to me 😅

Yes, this is weird. A related issue is #125.
I wonder if it will popup error if we don't remove the cache files. If so, we should set bang as true by default.

Issue #143, opened 2023-08-28
wrong CMakePresets.json file path


AbdelghaniDr:
**Bug description**
When there is a `CMakeUserPresets.json` file present in the project's root directory that includes another `CMakePresets.json` file from another directory (a child directory in the build folder for example), if the path is **absolute**, the plugin does not compose the correct path but just concatenates the two strings representing the paths, i.e, _the root project directory absolute path_ and the _absolute path for the included `CMakePresets.json` file_ and we end up with a string like **/home/user/projects/project1//home/user/projects/project1/build/CMakeUserPresets.json** which is not correct and causes a `can't open file` error when we call `CMakeSelectBuildPreset`.  

**Steps to reproduce**
1- directory structure:
```
 project1/
     CMakeUserPresets.json
     CMakeLists.txt
     main.cpp
     build/
        CMakePresets.json
```
2- content of the file CMakeUserPresets.json:
the file is auto generated by conan package manager
```
{
    "version": 4,
    "vendor": {
        "conan": {}
    },
    "include": [
        "/home/user/projects/project1/build/CMakePresets.json". <- note the absolute path here
    ]
}

```

**Expected behavior**
check if the paths are _absolute_ or _relative_ and compose the correct final path

**Environment**

- OS: Centos Linux

Civitasv:
Please also post the stacktrace.

Civitasv:
I've pushed a fix. Can you check it?

AbdelghaniDr:
it looks like it's fixed now, the plugin can open the included `CMakePresets.json` correctly, thanks

PR #152, opened 2023-09-12
Minor fixes - reading cmakepresets includes, add option to remove build preset and correct sourcedir on preset

ErezAmihud:
Hey, this merge request contains 3 little fixes/features:
1. If cmakepreset has an include for a file, we should read the file as a relative path to the presets file
2. add option of `buildPreset=None` in case  the user wants to not use a build preset anymore
3. Changed the sourceDir expantion on presets.json to be "." instead of the supplied cwd. The reason is the sourceDir is relative to the presets file. And when we invoke cmake we already use the new cwd, so the source dir is the current dir. (Without this change, using `CMakeRun` with a preset file with another cwd does not work.

### DIFF
--- a/lua/cmake-tools/init.lua
+++ b/lua/cmake-tools/init.lua
@@ -952,11 +952,10 @@ function cmake.select_build_preset(callback)
   local presets_file = presets.check(config.cwd)
   if presets_file then
     local build_preset_names = presets.parse("buildPresets", { include_hidden = false }, config.cwd)
-
     local build_presets =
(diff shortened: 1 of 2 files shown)


Civitasv:
> 1. If cmakepreset has an include for a file, we should read the file as a relative path to the presets file

The reason we use absolute path is #143.

ErezAmihud:
> > 1. If cmakepreset has an include for a file, we should read the file as a relative path to the presets file
> 
> The reason we use absolute path is #143.

Ok, I added an `if` for if the include path is absolute or relative

Issue #159, opened 2023-09-26
Wrong arguments passed to Overseer


lucas-yotsui:
**Bug description**
When I try to use Overseer as my executor it fails, apparently for passing the wrong cmake executable. here's the output I get in notify:

![error](https://i.imgur.com/bqS5bUC.png)

From this error I assume that the command passed to Overseer is wrong, since as you can see on the screenshot it is calling '  & /snap/bin/cmake' instead of just '/snap/bin/cmake'. The question now is: Is it something I configured wrong or is it an issue with the extension?

**Steps to reproduce**
here's my setup for the extension:

```lua
cmake.setup({
    cmake_command = '/snap/bin/cmake',
    cmake_build_directory = "build", -- this is used to specify generate directory for cmake, allows macro expansion
    cmake_kits_path = "/home/lucas-yotsui/.local/share/CMakeTools/cmake-tools-kits.json",
    cmake_executor = {               -- executor to use
        name = "overseer",           -- name of the executor
        opts = {
            new_task_opts = {},      -- options to pass into the `overseer.new_task` command
            on_new_task = function(task)
                task:subscribe("on_complete", LaunchEvent)
            end,
        },
    },
    cmake_terminal = {
        name = "terminal",
        opts = {
            name = "Terminal",
            prefix_name = "[CMakeTools]: ", -- This must be included and must be unique, otherwise the terminals will not work. Do not use a simple spacebar " ", or any generic name
            split_direction = "horizontal", -- "horizontal", "vertical"
            split_size = 11,

            -- Window handling
            single_terminal_per_instance = true,  -- Single viewport, multiple windows
            single_terminal_per_tab = true,       -- Single viewport per tab
            keep_terminal_static_location = true, -- Static location of the viewport if avialable

            -- Running Tasks
            start_insert_in_launch_task = true, -- If you want to enter terminal with :startinsert upon using :CMakeRun
            start_insert_in_other_tasks = true, -- If you want to enter terminal with :startinsert upon launching all other cmake tasks in the terminal. Generally set as false
            focus_on_main_terminal = true,      -- Focus on cmake terminal when cmake task is launched. Only used if executor is terminal.
            focus_on_launch_terminal = true,    -- Focus on cmake launch terminal when executable target in launched.
        },
    },
    cmake_dap_configuration = { -- debug settings for cmake
        name = "cpp",
        type = "lldb",
        request = "launch",
        stopOnEntry = false,
        runInTerminal = true,
[...]

Issue #158, opened 2023-09-19
CMakeRun not works on windows 


zuoxinyu:
**Bug description**
I'm using windows 11 and powershell, `:CMakeRun` generates command like:
```
cmd /C "cd "C:/Users/username/workspace/project/build" && C:/Users/username/workspace/project/build\bin/myapp.exe"
```
which leads a syntax error.

**Steps to reproduce**

<details><summary>Minimal configuration</summary>

    require("cmake-tools").setup {
        cmake_command = "cmake",                                           -- this is used to specify cmake command path
        cmake_regenerate_on_save = true,                                   -- auto generate when save CMakeLists.txt
        cmake_generate_options = { "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON" }, -- this will be passed when invoke `CMakeGenerate`
        cmake_build_options = {},                                          -- this will be passed when invoke `CMakeBuild`
        cmake_build_directory = "build",                                   -- this is used to specify generate directory for cmake
        cmake_build_directory_prefix = "cmake_build_",                     -- when cmake_build_directory is set to "", this option will be activated
        cmake_soft_link_compile_commands = true,                           -- this will automatically make a soft link from compile commands file to project root dir
        cmake_compile_commands_from_lsp = false,                           -- this will automatically set compile commands file location using lsp, to use it, please set `cmake_soft_link_compile_commands` to false
        cmake_kits_path = home .. '/.cmake-kits.json',                     -- this is used to specify global cmake kits path, see CMakeKits for detailed usage
        cmake_variants_message = {
            short = { show = true },                                       -- whether to show short message
            long = { show = true, max_length = 40 },                       -- whether to show long message
        },
        cmake_dap_configuration = {
            name = "cpp",
            type = "codelldb",
            request = "launch",
            stopOnEntry = false,
            runInTerminal = true,
            console = "integratedTerminal",
        },
    }

</details>

**Expected behavior**

`CMakeRun` executes selected binary correctly
**Screenshots**

**Environment**

- OS:  Windows 11
- Desktop environment:  powershell 7
- Plugins commit hash:  latest for now

**Additional context**
<!-- Add any other context about the problem here. --

rohit-kumar-j:
@zuoxinyu , if the issue is not yet fixed on master, you can use my branch temporarily as I use and support windows:

```lua
return { --lazy.nvim config
  "rohit-kumar-j/cmake-tools.nvim",
  branch = "master",
  config = function()
[...]

Issue #151, opened 2023-09-08
Toggle cmake_terminal


athanase:
**Describe the problem or limitation you are having**

While developing, I regularly toggle between the build terminal (in the quickfix list for this plugin) in a bottom window and also the launch terminal (here the [CMakeTools] terminal) in the same area. I currently use overseer which has a window displaying all tasks (build, run, ...) and [edgy](https://github.com/folke/edgy.nvim) to either have nothing open, or the quickfix window, or the tasklist.

I am most certainly missing something but I cannot find a way to toggle the `cmake_terminal` and also close it easily ?

**Describe the solution you'd like**

Preferred solution would be to have a way to hide/unhide/close the `cmake_terminal` easily ?
Or maybe have different execute strategy for the target runner (e.g. overseer or toggleterm instead of cmake_terminal) ? 

**Describe alternatives you've considered**

I am currently able to do this when I use overseer as it can be used to build or launch programs, and I can view it in the same tasklist pane. 

**Additional context**

Awesome plugin. Really waiting for the integration for the Test Framework integration.

rohit-kumar-j:
Related #145

rohit-kumar-j:
> **Describe the problem or limitation you are having**
> 
> While developing, I regularly toggle between the build terminal (in the quickfix list for this plugin) in a bottom window and also the launch terminal (here the [CMakeTools] terminal) in the same area. I currently use overseer which has a window displaying all tasks (build, run, ...) and [edgy](https://github.com/folke/edgy.nvim) to either have nothing open, or the quickfix window, or the tasklist.

Yes, currently there is no extensive terminal window handling for the `terminal executor` it can only be opened by running a `:CMake<...>` command. I think support for this can eventually be added.(`:CMakeOpen` and `:CMakeClose` support rudimentary open-close logic, but only for current terminal, not all open cmake terminals)

> I am most certainly missing something but I cannot find a way to toggle the `cmake_terminal` and also close it easily ?
>
> **Describe the solution you'd like**
> 
> Preferred solution would be to have a way to hide/unhide/close the `cmake_terminal` easily ? Or maybe have different execute strategy for the target runner (e.g. overseer or toggleterm instead of cmake_terminal) ?

[...]

Issue #145, opened 2023-08-31
support overseer for terminal handling


hfn92:
**Describe the problem or limitation you are having**
I find the terminal a bit intrusive. I'd prefer it to only show up when I want to.

**Describe the solution you'd like**
I wanted to see I if can adapt it to optionally run in a toggleable floating window. But I saw the new executor feature which supports overseer. It seems this only works for quickfix tough. Would it be possible to support this for the terminal as well. I think this would satisfy my use-case and even more.

rohit-kumar-j:
~How dare you... I feel attacked~ 😂 JK

Me and @Civitasv are currently developing the experimental terminal.

I share your sentiment for terminal, I think the existing terminal buffer options can have better windowing functionality, finetuning and setup instructions in the docs.

Do you suggest a togglable floating window like in [toggleterm](https://github.com/akinsho/toggleterm.nvim)?

I haven't dabbled with overseer at all, could give me an example project/config/setup/set of commands for overseer for this?

hfn92:
@rohit-kumar-j feel free to check out my fork here https://github.com/hfn92/cmake-tools.nvim/tree/overseer_terminal

This runs every cmake-tools command through overseer.  Overseer will show a notification if a command fails. I then open the output in a floating window to look at it. 

I use following keybindings
```lua
      ["<leader>of"] = { "<CMD> OverseerQuickAction open float<CR>", "Overseet open float " },
      ["<leader>oh"] = { "<CMD> OverseerQuickAction open hsplit<CR>", "Overseet open hsplit" },
      ["<leader>ov"] = { "<CMD> OverseerQuickAction open vsplit<CR>", "Overseet open vsplit" },
      ["<leader>oq"] = { "<CMD> OverseerQuickAction open output in quickfix<CR>", "Overseet open quickfix" },
      ["<leader>oo"] = { "<CMD> OverseerToggle<CR>", "Overseet toggle" },

```

rohit-kumar-j:
@hfn92 , I haven't fully dived deep into overseer, but does overseer have an inbuilt terminal mode?

Civitasv:
Thanks for @rohit-kumar-j review of overseer.nvim. I also have a quick look for this plugin. Here is my conclusion.

[...]

PR #160, opened 2023-10-01
Temp: Disable envScript in Overseer

rohit-kumar-j:
Temporarily fixing #158 and #159 by disabling envScript for overseer in preparation for #145 and #151


Civitasv:
Thanks!

PR #170, opened 2023-11-10
Command CMakeStop now stops terminal processes as well

Brynnjolf:
I noticed there isn't a simple way to cancel a running target, since the terminal is unmodified. This can be annoying when doing gui development as sometimes the exec loop is just running and you have no way to stop it.

I extended the `cmake.stop` function to stop any running terminals as well as executors.

Now when you run `:CmakeStop` it stops the running terminal too.


Civitasv:
Thanks!

PR #171, opened 2023-11-14
Fix: Re-Generate CMake only after selection of cwd has finished

jla2000:
Since 56b4c34 cwd selection is broken. Since this cwd selection is async, we should only regenerate when selection is finished. Otherwise the selection window will lose focus and the old cwd will get regenerated.

### DIFF
--- a/lua/cmake-tools/init.lua
+++ b/lua/cmake-tools/init.lua
@@ -1440,9 +1440,9 @@ function cmake.select_cwd(cwd_path)
         --if new_path:is_dir() then
         config.cwd = vim.fn.resolve(input)
         --	end
+        cmake.generate({ bang = false, fargs = {} }, nil)
       end)
     )
-    cmake.generate({ bang = false, fargs = {} }, nil)
   elseif cwd_path.args then
     config.cwd = vim.fn.resolve(cwd_path.args)
     cmake.generate({ bang = false, fargs = {} }, nil)


jla2000:
@Civitasv also would it be possible to add an option to disable this automatic regenerating? I find it quite annoying for projects that do a lot of operations in cmake, then you would have to wait lots of time or cancel it again with `:CMakeStop`

Civitasv:
> would it be possible to add an option to disable this automatic regenerating?

We already have this option: `cmake_regenerate_on_save`, forget to add it to docs.

jla2000:
> > would it be possible to add an option to disable this automatic regenerating?
> 
> We already have this option: `cmake_regenerate_on_save`, forget to add it to docs.

@Civitasv But actually we do not check it in the code? When setting this option to false, for me it will still regenerate.. There is also no check in the code itself, we just call regenerate directly.

Civitasv:
I understand your need, but I think if you change the current working directory, it is natural to regenerate.

jla2000:
Actually I run very often in the case that cmake-tools has configured a wrong directory when being initialized, so I would then need to regenerate every time I open the project again. Sometimes it even has v:null as cwd selected 🤔 but this gets fixed by selecting and regenerating a few times. I did not yet have time to debug it yet though.

Civitasv:
> Sometimes it even has v:null as cwd selected

Interesting, this must be a bug, can you draft an issue when you have time?

PR #178, opened 2023-12-11
fix: fix nil indexing if cmake-kits is missing

alfaix:
If `cmake-kits.json` is not there, line 27 causes nil indexing error


alfaix:
just realized this is a duplicate :sweat_smile: Can confirm I have the same bug though as #177

PR #185, opened 2023-12-17
Option to review command executed by :CMakeRun

minego:
Added an option that disables sending the new line character when using `:CMakeRun` to give the user a chance to review and/or modify the command before it is executed.


Civitasv:
Please fix the format.

PR #189, opened 2024-01-09
Fix: build dir for preset with inherited presets without configurePresets

lceWolf:
A (build) preset might inherit from a preset without a configurePresets field set.
This causes p_preset to be nil when entering `helper`. The check for `not p_preset` does
not catch the unconditional access to `p_preset.name` resulting in an access to a nil variable

returning an empty string for a nil preset with the addtional nil check for the `ppreset` should
prevent the nil access.

The second commit is optional, but restructures the helper function a bit


Civitasv:
Thanks for you contrib!

PR #207, opened 2024-04-13
Add support for *BSD

kareigu:
Adds support for *BSD systems by using the unix session directory structure.
Without this the plugin will crash on the FreeBSD install I'm using.

I wonder if it would make sense to have a generic `isunix` bool and use that in this check's place to handle more platforms that should be using the same directory structure.
Haven't tested if this has the same behaviour on Linux but at least on my FreeBSD machine `vim.fn.has("unix")` will return 1 as well, so that could be used as a catch all.


Civitasv:
> Haven't tested if this has the same behaviour on Linux but at least on my FreeBSD machine `vim.fn.has("unix")` will return 1 as well, so that could be used as a catch all.

It should return 1 on mac, linux and bsd, I think.

PR #210, opened 2024-04-16
Add option to not pass the '--presets' option to the cmake command

Townk:
## Changes Description

On CMake projects that use Conan as their package manager, there are two presets that are created with the same name. This always causes any `cmake` command issued by CMakeTools to fail. On those scenarios, the project usually specify the CMake options (including the preset location) as an argument to the `cmake` command.

This commit adds the option to not pass the `--preset` argument to the `cmake` command.

### DIFF
--- a/lua/cmake-tools/init.lua
+++ b/lua/cmake-tools/init.lua
@@ -83,7 +83,7 @@ function cmake.generate(opt, callback)
   -- if exists presets, preset include all info that cmake
   -- needed to execute, so we don't use cmake-kits.json and
   -- cmake-variants.[json|yaml] event they exist.
-  local presets_file = presets.check(config.cwd)
+  local presets_file = config.base_settings.use_preset and presets.check(config.cwd)
   if presets_file and not config.configure_preset then
     -- this will also set value for build type from preset.
     -- default to be "Debug"
@@ -242,7 +242,7 @@ function cmake.build(opt, callback)
   end
 
   local args
-  local presets_file = presets.check(config.cwd)
+  local presets_file = config.base_settings.use_preset and presets.check(config.cwd)
 
   if presets_file and config.build_preset then
     args = { "--build", "--preset", config.build_preset } -- preset don't need define build dir.
@@ -1238,7 +1238,7 @@ function cmake.create_regenerate_on_save_autocmd()
     table.insert(pattern, ss)
   end
 
-  local presets_file = presets.check(config.cwd)
+  local presets_file = config.base_settings.use_preset and presets.check(config.cwd)
   if presets_file then
     for _, item in ipairs({
       "CMakePresets.json",
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ local osys = require("cmake-tools.osys")
 require("cmake-tools").setup {
   cmake_command = "cmake", -- this is used to specify cmake command path
   ctest_command = "ctest", -- this is used to specify ctest command path
+  cmake_use_preset = true,
   cmake_regenerate_on_save = true, -- auto generate when save CMakeLists.txt
   cmake_generate_options = { "-DCMAKE_EXPORT_COMPILE_COMMANDS=1" }, -- this will be passed when invoke `CMakeGenerate`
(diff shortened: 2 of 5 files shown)


Townk:
I think I should re-word this PR, is not really about conan. The problem happens when you use CMake older than 3.23 AND conan. On that case, conan generated presets won't work with CMake. During compilation Conan actually logs that if you're using CMake < 3.23 you should not use presets, and that is the reason why I want an option to not pass --preset to cmake commands

Townk:
Just got time to gather more information regarding my reason for having this pull request.

First, about what I mentioned before, here is the relevant part of a Conan build log:

```
Calling generate()
conanfile.py (mypackage/0.1.0): Generators folder: /path/to/my/cmake/output/dir
conanfile.py (mypackage/0.1.0): CMakeToolchain generated: conan_toolchain.cmake
conanfile.py (mypackage/0.1.0): Preset 'conan-release' added to CMakePresets.json. Invoke it manually using 'cmake --preset conan-release' if using CMake>=3.23
conanfile.py (mypackage/0.1.0): If your CMake version is not compatible with CMakePresets (<3.23) call cmake like: 'cmake <path> -G "Unix Makefiles" -DCMAKE_TOOLCHAIN_FILE=/path/to/my/cmake/output/dir/conan_toolchain.cmake -DCMAKE_POLICY_DEFAULT_CMP0091=NEW -DCMAKE_BUILD_TYPE=Release'
conanfile.py (mypackage/0.1.0): CMakeToolchain generated: CMakePresets.json
conanfile.py (mypackage/0.1.0): CMakeToolchain generated: ../../../../CMakeUserPresets.json 
```

With that out of the way, I thought a bit more about why I want this pull request to go in. Of course, the main reason is that I want `cmake-tools` to work on my system, but besides that, I believe that making `cmake-tools` always use the `--preset` argument makes the plugin not as flexible as it could be. CMake itself does not imply `--preset` by default, so why should `cmake-tools` do it?

I'm not sure if this clarifies my reasons behind this pull request, but please let me know if you want further details about it.

I'm working on a second patch that will have this change rebased with the current `master`, and it should be ready around this week.

Civitasv:
I understand. I think we can add an option `use_presets`. It receive a function as its value, if the return value of this function is `true`, we use presets when presets file exists. It's more flexible than current and can solve this problem. What do you think?

Townk:
That is perfect, I would suggest the option to be a function _or_ a boolean.

Do you want me to make this change in this PR?

Civitasv:
> function _or_ a boolean.

Yeah, this is better.

> Do you want me to make this change in this PR?

Yes.

Townk:
Sorry for the delay here. I rebased the change and was able to simplify a bit. I made sure to let the `use_presets` option turned on by default to no break existing users.

lua/cmake-tools/init.lua
```diff
@@ -83,7 +83,7 @@ function cmake.generate(opt, callback)
   -- if exists presets, preset include all info that cmake
   -- needed to execute, so we don't use cmake-kits.json and
   -- cmake-variants.[json|yaml] event they exist.
-  local presets_file = presets.check(config.cwd)
+  local presets_file = config.base_settings.use_preset and presets.check(config.cwd)
```
Civitasv, line 86: It should be `use_presets`?

Townk: I was trying to follow the other variable names like `configure_preset`, `build_preset`, `select_configure_preset`, `configure_preset_names`, `format_preset_name`, etc.

And since you cannot use more than 1 preset at a time, I also thought that `use_preset` would be ok.

In all honesty, I have no preference for the singular or plural version of it. Do you want me to change everything for the "plural" version?

Civitasv: I think `use_preset` is better. But the problem is you mixed the use with `use_presets` and `use_preset`.

Townk: oh, I did not realize that 🤦🏼 

Will fix in a bit

Townk: Done, it should all be `use_preset`.

PR #214, opened 2024-04-23
feat: add option to run all tests to run_test()

khris190:
just added a option to use run_test({runAll = true}) to skip selecting tests and run all of them

### DIFF
--- a/lua/cmake-tools/init.lua
+++ b/lua/cmake-tools/init.lua
@@ -1307,27 +1307,27 @@ function cmake.target_settings(opt)
 end
 
 function cmake.run_test(opt)
-  opt = opt or {}
   if utils.has_active_job(config.runner, config.executor) then
     return
   end
 
   local env = environment.get_build_environment(config, config.executor.name == "terminal")
-  if opt.runAll then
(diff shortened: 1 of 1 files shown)


Civitasv:
Thanks for your contribution! Please change `runAll` to `run_all`. Add add description to [all_commands](https://github.com/Civitasv/cmake-tools.nvim/blob/master/docs/all_commands.md).

khris190:
> Thanks for your contribution! Please change `runAll` to `run_all`. Add add description to [all_commands](https://github.com/Civitasv/cmake-tools.nvim/blob/master/docs/all_commands.md).

I just realised how opts work in user_command, what would be the correct way to implement it, I'm thinking about changing the way I did this, so I would add an option to vim.ui.select all_tests at the top of the list called "run all",
but i dont know how to deal with the user_command because what i really wanted was to add a lua script way of running all the tests for a keybind and i dont really know how to deal with the user_command args so that those 2 can coexist in a clean way.

khris190:
i changed the way i did this so it just shows new option to run all tests
screen

Civitasv:
Yeah, this is better.

PR #215, opened 2024-04-24
feat: add arguments to cmd while using terminal.lua

khris190:
as the title says this change appends args when using default runner like in toggleterm.lua

### DIFF
--- a/lua/cmake-tools/terminal.lua
+++ b/lua/cmake-tools/terminal.lua
@@ -556,7 +556,7 @@ function _terminal.run(cmd, env_script, env, args, cwd, opts)
   end
 
   -- Send final cmd to terminal
-  _terminal.send_data_to_terminal(buffer_idx, cmd, {
+  _terminal.send_data_to_terminal(buffer_idx, cmd .. " " .. table.concat(args, " "), {
     win_id = final_win_id,
     prefix = opts.prefix_name,
     split_direction = opts.split_direction,
--- a/lua/cmake-tools/terminal.lua
+++ b/lua/cmake-tools/terminal.lua
@@ -556,7 +556,7 @@ function _terminal.run(cmd, env_script, env, args, cwd, opts)
   end
 
   -- Send final cmd to terminal
-  _terminal.send_data_to_terminal(buffer_idx, cmd .. " " .. table.concat(args, " "), {
+  _terminal.send_data_to_terminal(buffer_idx, cmd, {
     win_id = final_win_id,
     prefix = opts.prefix_name,
     split_direction = opts.split_direction,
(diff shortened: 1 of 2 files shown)


khris190:
Frankly, it feels more like a fix than a feat, because it just doesn't work without it, but I didn't know if there was any deeper thought behind it.

Civitasv:
When executor or runner is `terminal`, `args` will always be empty. Cause their args will be initialized using `terminal.prepare_cmd_for_run` and `terminal.prepare_cmd_for_execute`.

khris190:
so i need to find out why they don't initialize for me, thanks for the info

khris190:
`:CMakeRunTest` nor `require("cmake-tools").run_test()` seems to have `terminal.prepare_cmd_for_run` or `terminal.prepare_cmd_for_execute` in execution path and it just executes `ctest` command without args for me when using terminal runner

khris190:
that would mean, we need to move a check for "terminal" runner to utils.run instead of cmake.run?

khris190:
Considering that there are 3 places where we execute utils.run are init.lua:580, init.lua:653, and ctest.lua:37
where the execution from ctest.lua has an error where it doesnt append args because the other 2 places have an `if runner.name = 'terminal` check that executes terminal.prepare_cmd_for_execute i think it would be beneficial to move that check to utils but it nor then ill propose the pull request where i add analogic if check in `ctest.lua`

khris190:
i'm looking at the code and it looks like prepare_cmd_for_run should be a terminal.run() functions responsibility and not a callers responsibility why... was it made this way

khris190:
okay, i just added the analogical `if else` block to the ones in `init.lua` (i confused the files in commit desc).
I wanted to move those if else blocks to utils or even move this kind of handling to respective runners and executors bacuse this code feels like hydra and it doesn't make sense to make callers do runner specyfic checks and execution, in my opinion runners should deal with those things themselfs, but i didn't really understand why there is global full_cmd and calling callback before terminal cmd execution instead of in the on_exit callback like other runners so i didnt want to break anything, and most important i didnt want to create different logic for `run` and `execute`

Civitasv:
There is a global `full_cmd` because currently our terminal implementation does not support invoke a callback function after a specific cmd, for example, if there is a invoke chain `cmake.build(cb)->cmake.run(cb)`, which corresponds to `cmake --build . -> cb() -> ./exe -> cb()`, we don't know when `cmake --build .` will be finished. So we first initialize full_cmd to `cmake --build . && ./exe` then call `cb()`, does it make sense?

khris190:
i don't know where to put it so i'll write it here, i could try to implement on_exit in :term by pulling 
`ps --pid [child_process_id]` for unix and `tasklist /v /fi "PID eq [child_process_id]"` for windows it wouldn't be optimal but it doesnt seem possible any other way and not having this introduces complexity

PR #217, opened 2024-04-26
feat: add option to pass additional arguments to ctest

khris190:
so yea, this change lets us use `:CMakeRunTest --output-on-failure` or `cmake.run_test({args="--output-on-failure"})` and all the other flags that people might want or need, i added it because without  `--output-on-failure` flag there is no way of knowing what test in a single file failed like here:
without
here is the output with the flag
with
i thought about adding it as a default flag but it felt wrong and restrictive to the users.


khris190:
i could change it to a const variable to be set in config if You would like it more in that form

Civitasv:
Please fix the format.

khris190:
sorry about that, i'll be checking the style from now on.

PR #219, opened 2024-04-28
change terminal.lua to handle on_exit

khris190:
this one might require some changes so take Your time reviewing it because i might have broke some option that i didnt even know existed
- the terminal doesnt run in interactive mode now, to restore it we could add `&& ${SHELL}-bash` to the end of the command but then i feel like toggleterm should get this option too 
- i changed `has_active_job` because while trying out the "ps command pulling" method for on_exit i saw that bash has about 50-150ms downtime when running compound commands with `&&`, `;` or `||` when it just says that terminal doesnt have any children and in this implementation there is no need to check for children
- about the `prepare_cmd_for_execute` and `prepare_cmd_for_run` i feel like we should move all runners to the directory and create utils of them or even go for inheritence and create a base_runner that will contain those commands and run them in `utils.run` and `utils.execute`
- i also feel like `wrap_call` should be depricated and changed into `cmd_prexif` and add `cmd_postfix`, but that's just my feeling


Civitasv:
https://github.com/Civitasv/cmake-tools.nvim/issues/145#issuecomment-1713694078

khris190:
that makes sense, i could try to implement it with screen command but it wouldn't work on windows without cygwin

khris190:
or and i'm not proud of this idea but w/e
We could just write a lockfile when the command is run and add to end of every command `; rm {lockfile}` with a coroutine checking if that file exists to execute on_exit with `defer_fn`

khris190:
i know You probably thought of most of the options but it just doesn't let me sleep

Civitasv:
> or and i'm not proud of this idea but w/e We could just write a lockfile when the command is run and add to end of every command `; rm {lockfile}` with a coroutine checking if that file exists to execute on_exit with `defer_fn`

This maybe feasible.

Civitasv:
> i know You probably thought of most of the options but it just doesn't let me sleep

Don'r rush, take your time.

PR #220, opened 2024-04-30
feat: initial testing of lockfile and coroutine onexit implementation

khris190:
so this is initial version of lockfile on_exit implementation, it seems to work i just want to share the progress
TODO:

- i'll move lock_file functions to utils and parametrize them
- remove dump from 3rd line
- remove temporary sleep 5 command
- remove special treatment of that full_cmd and implement on_exit properly
- tests

if i'm missing something then please tell me

### DIFF
--- a/lua/cmake-tools/terminal.lua
+++ b/lua/cmake-tools/terminal.lua
@@ -549,15 +549,16 @@ end
 
 ---@return string
 local get_lock_file_path = function()
-  return utils.get_tmp_file_path("commandRunning.lock")
+  return get_tmp_file_path("command_running.lock")
 end
 
+---@return string
 local get_last_exit_code_file_path = function()
-  return utils.get_tmp_file_path("lastCmdExitCode.txt")
+  return get_tmp_file_path("last_cmd_exit_code")
 end
 
-local create_tmp_lock_file = function()
-  utils.create_tmp_file("commandRunning.lock")
+local create_lock_file = function()
+  create_tmp_file("command_running.lock")
 end
 
 ---creates command that handles all of our post command stuff for on_exit handling
--- a/lua/cmake-tools/terminal.lua
+++ b/lua/cmake-tools/terminal.lua
@@ -550,7 +550,17 @@ local file_exists = function(name)
   end
 end
 local on_exit_coroutine
-function _terminal.run(cmd, env_script, env, args, cwd, opts)
+
+---
+---@param cmd string command that will be run in the terminal
+---@param env_script any
+---@param env any
+---@param args string[]
+---@param cwd any
+---@param opts any
+---@param on_exit function|nil function to be called on exit
+---@param on_output any !unused here
+function _terminal.run(cmd, env_script, env, args, cwd, opts, on_exit, on_output)
   local prefix = opts.prefix_name -- [CMakeTools]
   create_tmp_lock_file()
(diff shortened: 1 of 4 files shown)


Civitasv:
Seems right to me.

khris190:
i added hardcoded 0 to on_exit() in terminal for now, but i'm thinking that we could store command exit code in another file so we could get it after

khris190:
also why don't runners have any callbacks?

Civitasv:
Executors are used for cmake-rated commands, including generate, build, install, clean, etc. Runners are used for running executables. Now the end of this plugin is Runner.

khris190:
i think it might be complete. i'll use it a bit so i'm sure everything works as intended

Civitasv:
> i'm also wondering if calling those files "tmp" is okay.

I think we should call it `cmake_tools_lockfile`

lua/cmake-tools/utils.lua
```diff
@@ -61,6 +61,22 @@ function utils.file_exists(path)
...
+---@return string
+function utils.get_tmp_file_path(file_name)
+  return utils.get_tmp_dir() .. "/" .. file_name
+end
+
+function utils.create_tmp_file(file_name)
+  utils.mkdir(utils.get_tmp_dir())
+  local tmp_file = io.open(utils.get_tmp_file_path(file_name), "w")
+  tmp_file:close()
+end
```
Civitasv, line 79: These functions are specific to `terminal`, should be moved to there.

lua/cmake-tools/terminal.lua
```diff
@@ -611,5 +668,16 @@ end
 function _terminal.is_installed()
   return true
 end
-
+--- Handle when a terminal process exists
+---@param opts any
+---@param on_exit function|nil function to be executed on exit
+---@param close_on_exit boolean close the terminal on exit
+function _terminal.__handle_exit(opts, on_exit, close_on_exit)
```
Civitasv, line 675: why not just name it to `handle_exit`?

lua/cmake-tools/terminal.lua
```diff
@@ -527,9 +530,52 @@ function _terminal.prepare_cmd_for_execute(cmd, env, args)
...
-  local prefix = opts.prefix_name -- [CMakeTools]
+local get_lock_file_path = function()
+  return utils.get_tmp_file_path("commandRunning.lock")
+end
+
+local get_last_exit_code_file_path = function()
+  return utils.get_tmp_file_path("lastCmdExitCode.txt")
+end
+
+local create_tmp_lock_file = function()
```
Civitasv, line 541: create_lock_file is okay

lua/cmake-tools/terminal.lua
khris190: i copied it from toggleterm without thinking, i now see that it doesn't work with our naming scheme

lua/cmake-tools/utils.lua
khris190: my idea was that the tmp directory handling is not specyfic to terminal, it is just that only terminal uses it for now, but if we ever need any other tmp file stored we could use the same functions

Civitasv: I still think they are too specific to be in utils.

Civitasv:
LGTM. Thanks for your fantastic work!

PR #221, opened 2024-05-01
feat: add single_terminal_per_instance option for toggleterm

### LINKED ISSUE #209: Toggleterm executor opens new tab instead of rewriting existing one
My config is
```lua
return {
  "Civitasv/cmake-tools.nvim",
  config = function()
      cmake_executor = {
        name = "toggleterm",
        opts = {
          direction = "tab", -- 'vertical' | 'horizontal' | 'tab' | 'float'
          close_on_exit = false, -- whether close the terminal when exit
          auto_scroll = true, -- whether auto scroll to the bottom
        },
      },
    }
  end,
}
```
If i rerun `CMakeGenerate`, it creates new tab. I am not sure this is optimal behaviour. Can it just clear existing one and write cmake output into it. 

Following is more feature request, than bug, but it seeml logical to have one toggleterm output for each preset\variant\profile.

TheRealLorenz:
As the name implies, uses a single instance of toggleterm by closing the already opened one.

Closes #209

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ require("cmake-tools").setup {
         direction = "float", -- 'vertical' | 'horizontal' | 'tab' | 'float'
         close_on_exit = false, -- whether close the terminal when exit
         auto_scroll = true, -- whether auto scroll to the bottom
-        single_terminal_per_instance = false, -- single instance, autocloses the opend one, if present
+        single_terminal_per_instance = false, -- single instance, autocloses the opened one, if present
       },
       overseer = {
         new_task_opts = {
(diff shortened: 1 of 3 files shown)


lua/cmake-tools/toggleterm.lua
```diff
@@ -24,6 +24,9 @@ end
 
 function _toggleterm.run(cmd, env_script, env, args, cwd, opts, on_exit, on_output)
   _toggleterm.cmd = cmd .. " " .. table.concat(args, " ")
+  if opts.single_terminal_per_instance and _toggleterm.term then
+    _toggleterm.term:close()
+  end
```
Civitasv, line 29: I think we can make this as default.

README.md
```diff
@@ -114,6 +115,7 @@ require("cmake-tools").setup {
         direction = "float", -- 'vertical' | 'horizontal' | 'tab' | 'float'
         close_on_exit = false, -- whether close the terminal when exit
         auto_scroll = true, -- whether auto scroll to the bottom
+        single_terminal_per_instance = true, -- single instance, autocloses the opened one, if present
```
Civitasv, line 118: Maybe `singleton` is enough, what do you think?

TheRealLorenz: Sounds good to be, I was trying to be consistent with the terminal's settings

Civitasv:
Thanks for your contrib!

PR #223, opened 2024-05-03
Fix call in ctest

khris190:
i saw the "depricated" commit, doesnt deprication mean that its a heads up before it gets removed?


Civitasv:
Thanks! I forgot this.

khris190:
> Thanks! I forgot this.

yep, that's why the changes right now are great, not needing to do extra checks in every place we call run/execute

PR #224, opened 2024-05-03
fix: normalize paths in generating autocmd

zuoxinyu:
Fixes an path issue on Windows.


Civitasv:
Thanks for your contrib! I didn’t know these two functions before, maybe somewhere else should also use these two functions.

Civitasv:
vim.fs.joinpath is not avaliable with nvim 0.9.5, from what I search, it appears in 0.10.0, which I think would be not suitable to use it for now.

zuoxinyu:
Usually "/" is enough, but the pattern in autocmd may be an exception, and since `vim.fs` requires a higher version, we can roughly use "/" on unix or "\" on windows respectively.

PR #225, opened 2024-05-03
fix: command not ending when cmd doesnt exit successfully

khris190:
this part is quite important expecially for tests because we want to get the exit code and delete lock even if command exits with an error code for example: when tests are failing and tests are made to fail

### DIFF
--- a/lua/cmake-tools/terminal.lua
+++ b/lua/cmake-tools/terminal.lua
@@ -597,7 +597,7 @@ function _terminal.run(cmd, env_script, env, args, cwd, opts, on_exit, on_output
   end
 
   -- Send final cmd to terminal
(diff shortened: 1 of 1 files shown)


Civitasv:
Makes sense. I also removed the `return $EXITCODE` to make command shorter, is that ok?

image

khris190:
> Makes sense. I also removed the `return $EXITCODE` to make command shorter, is that ok?
> 
> image
> 

I mean depends, i had it there, so on the offchance that someone wanted that exit code, they could use it as our file handling wasn't there, My zsh shows me the exit code after the command so personally i just wanted to see what the command returned. But it doesn't really break anything  if we don't have it there

Civitasv:
ok, I understand.

Issue #213, opened 2024-04-23
Symlink compile_commands.json doesn't appear to work on windows


natecheadle:
**Bug description**
compile_commands.json is not being symlinked to the root directory of the source folder on windows. 

Running the same command that is being passed to vim.cmd works properly from the same CLI environment. 
cmake -E create_symlink {source} {destination}

This works fine on OSX. 

(I did not test my project on Linux)

**Steps to reproduce**
clone https://github.com/natecheadle/CppTemplate
mkdir out/Debug
nvim .
:CMakeGenerate

**Expected behavior**
expected compile_commands.json to be symlinked in the root source directory

**Environment**

- OS: Windows
- Plugins commit hash: a4cd0b3a2c8a166a54b36bc00579954426748959
- CLI: Launched from Powershell

**Work Around**
I was able to work around this issue by adding the following to my main CMakeLists.txt and disabling the symlink feature in this plugin. (This does exactly what this plugin is trying to do as far as I can tell)

```
execute_process(
    COMMAND ${CMAKE_COMMAND} -E create_symlink  
    ${CMAKE_BINARY_DIR}/compile_commands.json  
    ${CMAKE_SOURCE_DIR}/compile_commands.json  
)
```

khris190:
what do You get when you execute `print(vim.loop.cwd())` it might be an error with that, it should be your ${CMAKE_SOURCE_DIR}

natecheadle:
`:lua print(vim.loop.cwd())` returns my {CMAKE_SOURCE_DIR}. 

This is a different windows PC than what I originally reported the issue but it does still fail. On my other PC showing the issue it was a silent failure, now it does report an error in the function trying to symlink the file. The commit of CMake-Tools that I am using on this PC does look to be different though (8506c426d5b361dbaa7c61c0490a2589a9c555ce).

The following error is what I receive. 

```
eposIgnosiGameEngine.out^HuildMSVCRelease/compile_commands.json C:Usersec2(): Vim:E492: Not an editor command: atecsource
eposIgnosiGameEngine/compile_commands.json                                                                                                                              stack traceback:
        [C]: in function 'nvim_exec2'
        vim/_editor.lua:341: in function 'cmd'
        ...vim-data/lazy/cmake-tools.nvim/lua/cmake-tools/utils.lua:90: in function 'softlink'
[...]

PR #226, opened 2024-05-05
fix: transform path in utils.softlink()

khris190:
i had to swap `"` to `'` because we wrap paths in `"`, this fixes symlink not being created if the path contains the space character

possibly wil fix [issue213](https://github.com/Civitasv/cmake-tools.nvim/issues/213)


Civitasv:
👍

Issue #228, opened 2024-05-06
Build Directory Creation will Fail if it Requires a Recursive Create on Windows


natecheadle:
**Bug description**
During the generate step if your output directory requires a recursive create on Windows it will fail. In my example my output directory is ${CMAKE_SOURCE_DIR}/out/Debug/.

If the directory "out" exists the command succeeds, but if it does not the command will fail.

```
Error executing Lua callback: ...a/Local/nvim-data/lazy/plenary.nvim/lua/plenary/path.lua:505: We couldn't mkdir: C:\repos\CppTemplate\out/Debug
stack traceback:
        [C]: in function 'error'
        ...a/Local/nvim-data/lazy/plenary.nvim/lua/plenary/path.lua:505: in function 'mkdir'
        ...im-data/lazy/cmake-tools.nvim/lua/cmake-tools/config.lua:120: in function 'generate_build_directory'
        ...nvim-data/lazy/cmake-tools.nvim/lua/cmake-tools/init.lua:162: in function <...nvim-data/lazy/cmake-tools.nvim/lua/cmake-tools/init.lua:61>
```

**Steps to reproduce**
clone https://github.com/natecheadle/CppTemplate
nvim .
:CMakeGenerate

**Expected behavior**
I would expect no failure on the generate step.

**Environment**

- OS: Windows
- Plugins commit hash: 1c309e4

khris190:
if this works i'll link a pull request

natecheadle:
Changing the build directory as you describe works. 

This probably isn't a surprise based on what I assume your fix will be, but using CMakePresets to set the binary directory works without issues.

khris190:
i tried fixing it the most basic way and put handling of windows slashes in all paths in my personal TODO list

PR #230, opened 2024-05-07
feat: change default build path for windows

khris190:
i added a very basic override when config is run on windows, didn't really know how to adnotate it properly in readme.
this should fix: [issue228](https://github.com/Civitasv/cmake-tools.nvim/issues/228),
if there will be a lot of mixed slashes errors for windows users we could add an slash check and replace in `utils.transform_path`

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ require("cmake-tools").setup {
   --       ${kit}
   --       ${kitGenerator}
   --       ${variant:xx}
-  cmake_build_directory = "out/${variant:buildType}", -- this is used to specify generate directory for cmake, allows macro expansion, relative to vim.loop.cwd()
+  cmake_build_directory = "out/${variant:buildType}", -- this is used to specify generate directory for cmake, allows macro expansion, relative to vim.loop.cwd(). ON WINDOWS => "out\\${variant:buildType}"
   cmake_soft_link_compile_commands = true, -- this will automatically make a soft link from compile commands file to project root dir
   cmake_compile_commands_from_lsp = false, -- this will automatically set compile commands file location using lsp, to use it, please set `cmake_soft_link_compile_commands` to false
   cmake_kits_path = nil, -- this is used to specify global cmake kits path, see CMakeKits for detailed usage
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ The goal of this plugin is to offer a comprehensive, convenient, and powerful wo
 ## :balloon: Configuration
 
 ```lua
+local osys = require("cmake-tools.osys")
 require("cmake-tools").setup {
   cmake_command = "cmake", -- this is used to specify cmake command path
   ctest_command = "ctest", -- this is used to specify ctest command path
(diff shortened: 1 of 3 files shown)


Civitasv:
I don't think it's a decent solution. Maybe it's better to handle this error in the `Config:update_build_dir` function:

https://github.com/Civitasv/cmake-tools.nvim/blob/1c309e42bf0fe1ee79caa4de7b3cc39a170de382/lua/cmake-tools/config.lua#L61

lua/cmake-tools/const.lua
```diff
@@ -114,5 +114,8 @@ local const = {
   },
   cmake_virtual_text_support = true, -- Show the target related to current file using virtual text (at right corner)
 }
-
+local osys = require("cmake-tools.osys")
+if osys.iswin32 then
+  const.cmake_build_directory = "out\\${variant:buildType}"
```
Civitasv, line 119: If it is windows, and the path includes `/`, then replace `/` to `\\`.

Civitasv:
Or, we can change `cmake_build_directory` to accept function or string, and handle it in the `Config:update_build_dir` function.

khris190:
> Or, we can change `cmake_build_directory` to accept function or string, and handle it in the `Config:update_build_dir` function.

that makes a lot more sense

khris190:
i added an error so we get sth like this if someone passes something that isnt a string or a function returning a string

lua/cmake-tools/const.lua
khris190: i think we should possibly add that replace to utils.transform_path later if there will be more errors like this

Civitasv:
> i added an error so we get sth like this if someone passes something that isnt a string or a function returning a string 

ok, this is good.

README.md
```diff
@@ -34,7 +34,12 @@ require("cmake-tools").setup {
   --       ${kit}
   --       ${kitGenerator}
   --       ${variant:xx}
-  cmake_build_directory = "out/${variant:buildType}", -- this is used to specify generate directory for cmake, allows macro expansion, relative to vim.loop.cwd()
+  cmake_build_directory = function()
```
Civitasv, line 37: You forgot to add the `osys` module here.

README.md
```diff
@@ -34,7 +34,12 @@ require("cmake-tools").setup {
...
+    if osys.iswin32 then
+      return "out\\${variant:buildType}"
+    end
+    return "out/${variant:buildType}"
+  end, -- this is used to specify generate directory for cmake, allows macro expansion, can be a string or a function returning the string, relative to vim.loop.cwd().
```
Civitasv, line 42: change to relative to cwd, cause cwd can be changed (using `CMakeSelectCwd`).

khris190:
this one is ready i think

khris190:
w8 something broke

khris190:
i dont know how i tested this but we are calling that function using our config build directory as both params so we need both of them checked and executed

khris190:
it should be ready now

Civitasv:
Good work.

Issue #201, opened 2024-02-24
Usage/workflow questions


TheSunCat:
Hi! I set up this plugin with the terminal option, and it does its job. Thanks for maintaining it!

I believe I may be using it wrong though, as I have encountered several issues with it:

1. ~~it seems to always create a buffer called `*cmake-tools*` that shows up in my bar and gets saved as a file when I do `:wa`~~ This is tracked at #198 
2. it outputs to a terminal that opens at the bottom, but this terminal does not scroll down to the latest output, so I am always looking at the first build/generate I ever did
3. the terminal stays open and if I use a keybind to go to the next or previous buffer, it changes the terminal into the contents of the file I was going to.
4. sometimes the terminal CD's to the build directory, and then none of the commands work as they cannot find a CMakeLists.txt until I manually run `:cd ..`
5. the toggleterm mode starts printing errors if my program logs too much, and eventually fully freezes nvim (requiring force closing my terminal emulator)

Did I mess up my config or is this normal behavior?

Config:

<details>
<summary>cmake-tools config</summary>

```lua
require("cmake-tools").setup {
  cmake_command = "cmake", -- this is used to specify cmake command path
  ctest_command = "ctest", -- this is used to specify ctest command path
  cmake_regenerate_on_save = true, -- auto generate when save CMakeLists.txt
  cmake_generate_options = { "-DCMAKE_EXPORT_COMPILE_COMMANDS=1" }, -- this will be passed when invoke `CMakeGenerate`
  cmake_build_options = {}, -- this will be passed when invoke `CMakeBuild`
  -- support macro expansion:
  --       ${kit}
  --       ${kitGenerator}
  --       ${variant:xx}
  cmake_build_directory = "out/${variant:buildType}", -- this is used to specify generate directory for cmake, allows macro expansion, relative to vim.loop.cwd()
  cmake_soft_link_compile_commands = true, -- this will automatically make a soft link from compile commands file to project root dir
  cmake_compile_commands_from_lsp = false, -- this will automatically set compile commands file location using lsp, to use it, please set `cmake_soft_link_compile_commands` to false
  cmake_kits_path = nil, -- this is used to specify global cmake kits path, see CMakeKits for detailed usage
  cmake_variants_message = {
    short = { show = true }, -- whether to show short message
    long = { show = true, max_length = 40 }, -- whether to show long message
  },
  cmake_dap_configuration = { -- debug settings for cmake
    name = "cpp",
    type = "codelldb",
[...]

PR #234, opened 2024-05-19
fix(quickfix): auto_close_when_success false is ignored

luxstudio90:
When the runner/executor is set to `quickfix` and in the options `auto_close_when_success` is set to `false`, the `quickfix` window does not remain open.

This PR adds the missing check when success code = 0, auto close = false.

The issue is referenced also in the first bullet of https://github.com/Civitasv/cmake-tools.nvim/issues/201#issuecomment-2022290445

### DIFF
--- a/lua/cmake-tools/quickfix.lua
+++ b/lua/cmake-tools/quickfix.lua
@@ -69,13 +69,7 @@ function _quickfix.run(cmd, env_script, env, args, cwd, opts, on_exit, on_output
         _quickfix.scroll_to_bottom()
       end
       if code == 0 and opts.auto_close_when_success then
-        vim.defer_fn(function()
-          _quickfix.close(opts)
-        end, 100)
-      end
-      if code == 0 and not opts.auto_close_when_success then
-        _quickfix.show(opts)
-        _quickfix.scroll_to_bottom()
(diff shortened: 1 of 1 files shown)


lua/cmake-tools/quickfix.lua
```diff
@@ -73,6 +73,10 @@ function _quickfix.run(cmd, env_script, env, args, cwd, opts, on_exit, on_output
           _quickfix.close(opts)
         end, 100)
       end
+      if code == 0 and not opts.auto_close_when_success then
+        _quickfix.show(opts)
+        _quickfix.scroll_to_bottom()
```
Civitasv, line 78: When code equals 0, and `auto_close_when_success` is set to false, it should remain open. I think `_quickfix.show(opts)` is redundant, only `_quickfix.scroll_to_bottom()` is neccessary.

Civitasv: I can reproduce this with this config. But the pr cannot fix it. I think the reason is from `vim.defer_fn`.

Civitasv: I think we can remove this.

Civitasv:
Thanks

Issue #235, opened 2024-05-20
CMakeRun No Output on Windows


Skeeledon:
**Bug description**

The output from the program is not displayed.

**Steps to reproduce**
1. Generate and Build Program that output texts to cout
2. Call CMakeRun and see no text being outputted.

<details><summary>Minimal configuration</summary>
Just the CMake-Tools Plugin 
</details>

**Cause**
In `get_command_handling_on_exit` the line "echo $? > " seems to be the cause. Removing '>' seems to fix it. Though having a local work around forces Lazy to complain. Also windows does not use `rm`. Believe the its rmdir. Can probably do something like
 if osys.iswin32 then
    .. " && rmdir"
 else
    .. " && \\rm -f "
 end
 
 Though this gives a new error "Invalid switch - "cmake-tools-tmp"."

Jorissss:
Running into this issue with `rm` as well.

Jorissss:
This is what worked for me:
```lua
local get_command_handling_on_exit = function()
  -- Set commands to Windows or Unix version
  local echo_cmd = osys.iswin32 and "echo %errorlevel% " or "echo $? > "
  local rm_cmd = osys.iswin32 and " && cmd /C del /Q " or "&& \\rm -f "

  local exit_code_file_path = get_last_exit_code_file_path()
  local lock_file_path = get_lock_file_path()

  -- Normalize paths for Windows
  if osys.iswin32 then
    exit_code_file_path = exit_code_file_path:gsub("/", "\\")
    lock_file_path = lock_file_path:gsub("/", "\\")
  end

  return echo_cmd
    .. exit_code_file_path -- write exit code to file
    .. rm_cmd
    .. lock_file_path
end

```

khris190:
i need to create a VM for all of the windows problems do i just install nvim with chocolatey and test it on terminal and powershell?

Skeeledon:
Pretty much. I use Windows Terminal + PowerShell 7. Installed all my dependencies through choco.

Civitasv:
I think this should be fixed by #236. Closing for now.

PR #236, opened 2024-05-26
add windows-specific logic to runner code

Jorissss:
Fix for #235, adds some conditionals with windows-specific commands for `CMakeRun` command.


Jorissss:
Actually when we remove the `>` it does not write the exit code to the file. When we add it, it writes the program output to the exit code file...

Jorissss:
Seems `;` is not a valid command separator on windows, replaced with `&&`.

Issue #245, opened 2024-06-24
Suggestion: provide a stable callback-based lua API


alfaix:
Hey, thank you for all the work on this plugin. It really makes interacting with CMake much easier and I use it a lot :)

That said, I think the plugin can be improved by providing a stable API.

**Describe the problem or limitation you are having**
The plugin's API is difficult to use by another plugin.

**Describe the solution you'd like**
1. Declare which part of the lua API is considered stable
2. Follow the traditional `fn(is_ok, result_or_error)` callback signature in all lua API

**Describe alternatives you've considered**
Working with the API as-is. It's quite difficult, and here's an example:
I'm working on [this plugin](https://github.com/alfaix/neotest-gtest), and as part of it I need to list all executables in the CMake project.
To do that, I need to generate the CMake files first. Thus, I would have to call `cmake.generate()`, and then `cmake.get_targets`. However, there are 2 problems:
1. `cmake.generate` only calls the callback if the call is successful. This means that I cannot schedule another operation after it: once I do `cmake.generate()`, I have no way of knowing when it is done. I can pass the callback, but if the call fails, I'll never know. One option is to do timeouts, but then if CMake fails quickly (say, due to invalid `CMakeLists` file), the user would have to wait until the timeout (which can be very long for complex `CMakeLists`) before the interface would react.
2. I can't know if the API will change in the future without notice, as the only documented API is vim commands.

**Additional context**
If the proposed changes make sense to you (change callback signature, ensure callback is always called), I can take a crack at it in a PR.

alfaix:
Hey @Civitasv I was wondering if you would accept a PR with such changes (make all callbacks adhere to the same format, make sure they are always called even in case of command failure)?

Civitasv:
We actually have some getters in `init.lua`, which I think is stable, but I don't document them yet. See
https://github.com/Civitasv/cmake-tools.nvim/blob/5893fef535912baa4b3bae4dde550211a90e699d/lua/cmake-tools/init.lua#L1020



> Hey @Civitasv I was wondering if you would accept a PR with such changes (make all callbacks adhere to the same format, make sure they are always called even in case of command failure)?

I think this indeed is helpful. And I'm happy to accept this.

PR #246, opened 2024-06-27
feat!(API): add callback(result) to lua API

alfaix:
In this PR I add callback(result) as the last parameter to most API calls, and:
1. Make sure it is always called exactly once, be that in case of success or failure
2. Fix a bug with `Result` type: its `new` method was setting attributes on the class instead of the object
3. Fix a bug with `Types` type: `SUCCESS` was `1` when there are places in code that assume that it is 0
4. Add a couple of error messages

I tested this locally - to an extent, let me know if there is some kind of proper procedure to know it's working.

RESOLVES: #245

### DIFF
--- a/lua/cmake-tools/init.lua
+++ b/lua/cmake-tools/init.lua
@@ -81,10 +81,10 @@ end
 ---@return nil
 function cmake.generate(opt, callback)
   callback = callback or function() end
-  callback(Result:new(Types.SUCCESS, nil, nil))
   if check_active_job_and_notify(callback) then
     return
   end
+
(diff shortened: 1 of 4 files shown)


Civitasv:
LGTM. This is indeed improve the API quality, thanks for your work! I think you should also document it somewhere. Maybe create a API file in the docs folder.

tan-wei:
@Civitasv Does the API could be integrated with cmake-tools.nvim to help us to locate the gtest binaries? Currently, it is quite convenient to select a binary from neotest. It would be great if any snippets exist, thanks.

PR #250, opened 2024-08-26
Toggleterm: improve focusing and scrolling capabilities

deeedob:
Hey, I use these features all the time and maybe worth merging :)


lua/cmake-tools/const.lua
```diff
@@ -41,7 +41,10 @@ local const = {
       toggleterm = {
         direction = "float", -- 'vertical' | 'horizontal' | 'tab' | 'float'
         close_on_exit = false, -- whether close the terminal when exit
-        auto_scroll = true, -- whether auto scroll to the bottom
+        auto_scroll = true, -- auto scroll on new input
+        scroll_on_error = false, -- scroll to bottom on error
+        auto_focus = true, -- auto focus the terminal on activation
```
Civitasv, line 46: `scroll_on_error` makes sense only `auto_scroll` is false, right?

deeedob: Maybe some people like `auto_scroll`, it's still valid. For me its too aggressive (it's pushing the cursor to the end of the buffer on update)

Civitasv:
Thanks for your contribution!

PR #251, opened 2024-09-12
feat: add function to retrieve launch arguments

lceWolf:
The api allows setting the launch argument already, but it lacks an option to retrieve the currently set arguments. 
This commits adds the missing function.
This enables use cases like appending new arguments to or removing some from the previously set ones.


lceWolf:
Looks like I overlooked something. This function already exists [here](https://github.com/Civitasv/cmake-tools.nvim/blob/master/lua/cmake-tools/init.lua#L1049)

Issue #247, opened 2024-07-22
Automatic split size?


TheComet:
**Describe the problem or limitation you are having**
If I manually resize the terminal split, running ```:CMakeRun``` will resize it back to whatever ```split_size``` is set to.

**Describe the solution you'd like**
I'd like a way to configure cmake-tools to not resize my terminal window and use whatever it is currently set to.

lceWolf:
proposal implemented in #253

lceWolf:
implementation merged. Issue can be closed

PR #253, opened 2024-09-18
feat: add option to not resize the terminal (#247)

lceWolf:
I'm not sure if the option name is fitting. I am open to suggestions.

Open question: Should the `keep_terminal_size` option be `true` per default?


lceWolf:
renamed the option - I think this fits way better.

Still, should the default behaviour be changed to not automatically resize the terminal, or should it be kept opt-in?

Civitasv:
> should the default behaviour be changed to not automatically resize the terminal

I don't think so. It should be kept opt-in.

Civitasv:
Thanks for your contributing!

lceWolf:
I think issue #247 can be closed now @Civitasv

PR #254, opened 2024-09-20
feat: improve runner and executer notification handling

lceWolf:
With the new approach we get rid of the global states managed in the notifications table. Everything is capsulated within the created notification instance.
That also allows to fix an issue where a notification is dismissed while a build process is still ongoing and the spinner should be updated. Updating the spinner fails, because the notification timed out and got dismissed, resulting in a repeating "No Matching notification to replace found" error.
The new instance based approach tracks the shown notification and replaces any existing notification when the message is updated. If the notification times out and is dismissed, the instance will create a new notification automatically.

Additionally, the notification window is resized automatically to match the updated contents, if the displayed message changes

### DIFF
--- a/lua/cmake-tools/init.lua
+++ b/lua/cmake-tools/init.lua
@@ -34,6 +34,8 @@ function cmake.setup(values)
     const.cmake_runner.opts or {}
   )
 
+  require("cmake-tools.notification").setup(const.cmake_notifications)
+
   config = Config:new(const)
 
   -- auto reload previous session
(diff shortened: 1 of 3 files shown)


lceWolf:
The issue I ran into was as follows:
When building a project using all cores, neovim will eventually (partly) freeze and the notification displaying the build progress will not get updated before it reaches its timeout. When neovim unfreezes, the notification is dismissed before cmake-tools can refresh it. Now the `replace` option refers to an invalid notification id, resulting in a repeatedly shown error from vim-notify.

The same behaviour can be provoked by using noice and dismissing all notifications using `noice.dismiss`.

Issue #227, opened 2024-05-06
targets in buildPresets of CMakePresets.json not honored


pepe82sh:
**Bug description**
Running `CMakeBuild` on a build preset builds the `all` target, not the targets configured in the build preset.

**Steps to reproduce**
run `:CMakeSelectBuildPreset` and select a build preset that has `"targets": ["a", "b", "c"]`
run `:CMakeBuild`

<details><summary>Minimal configuration</summary>
`require("cmake-tools").setup{}`
</details>

**Expected behavior**

The targets listed in the build presets `target` section are being built in case there are targets defined in the build preset, as would be the case when running `cmake --build --preset MyPreset`

**Environment**

- OS: Ubuntu
- Desktop environment: None
- Plugins commit hash: 1c309e4

pepe82sh:
@lceWolf Unfortunately this still seems to artificially use "all" target when a build preset has been defined. So it appears that the refactoring isn't fixing this issue really.

pepe82sh:
[The code here in init.lua](https://github.com/Civitasv/cmake-tools.nvim/blob/a5d5bf0a93bb96085502334a5036fc054ad9cf33/lua/cmake-tools/init.lua#L356C1-L365C6) has some special handling for `all` that seems not necessary. I assume the idea was to omit the `--target` option when `all` is set.
This could be controversial, because someone may want to set `all` deliberately to override the default of the preset (not my use case).
The other alternative would be to define a `default` target that will explicitly omit the `--target` option and therefore use the CMake default (`all` or whatever the preset has defined).
Yet if someone has an actual `default` target defined it'd be an issue, but I've never seen that anywhere.

Civitasv:
I've fixed this. Use the latest commit.

PR #255, opened 2024-09-20
Presets refactoring

lceWolf:
This PR comes with an overhaul of the preset parsing / handling. 
It contains the followin changes:

### configure presets are now actual class instances with encapsulated data.
This is the enablement for the other features contained in the PR. 

### reduced access to disk 
As the configure preset data is now encapsulated in the class instance, it is not necessary to read the preset file(s) multiple times to access different parts of the preset. 
Previously, the most extreme case was determining the build_dir, which read the preset file for each inherited preset.
The new approach reads the file exactly once and parses the data into a hierarchy of preset instances.

### presets from CMakePrests.json honored (#227)
If a CMakeUserPreset.json is found, the old implementation did not look up neither the inherited presets nor non-hidden presets from a CMakePreset.json file.
The new implementation reads both files and combines them into one data structure. The values from the CMakeUserPreset.json take precedence.

### environment variable support
The new instance based approach allows building a environment table. This enables resolving environment variables (`$env{<var>}` and `$penv{<var>}`) for the environment table itself, but also other fields. 
MacroExpansion (including envVars) currently supported for the binaryDir, condition variables and CacheVariables.

### conditions 
This PR adds support for conditions (except for `matches` and `notMatches` due to lua's regex capabilities).
If a preset is disabled via its condition, it is removed from the list of available configure presets and automatically deselected, if it is the currently selected configure preset.
If a configure preset and thereby the corresponding, currently selected build preset becomes available again, the configure preset is automatically selected again (based on the currently selected build preset).

Also included is a config option to hide any build preset which refers to a disabled configure preset. This is not standard cmake behaviour, but I think it is quite nice to be able to hide build preset's whose configure presets can not be used anyways.
Currently, this option is added to the `base_settings` and saved per session. This allows defining a default in the config, but re-configure it for inidividual projects, if needed.

### misc
This PR prevents an error which occurs when `cmake --target clean` is issued if no cache is available. Now, the cache is only cleared, if it actually exists.

I moved the call to `generate` from `BufWritePre` to `BufWritePost` as this caused an error when a preset got disabled in the currently opened CMake[User]Preset.json file. `generate` would prompt the user to select another configure preset while the buffer is being written. This caused the preset selection to lose focus and become untargetable. Running `generate` after the buffer was written fixes this issue.

### DIFF
--- a/docs/cmake_presets.md
+++ b/docs/cmake_presets.md
@@ -12,4 +12,4 @@ Read more about CMake presets from [CMake docs](https://cmake.org/cmake/help/lat
 2. [Package Preset](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#package-preset) is not supported.
 3. [Workflow Preset](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#workflow-preset) is not supported.
 4. [Condition](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#condition) mostly supported. Types `matches` and `notMatches` currently not supported due to lua's differences in regex capabilities
-5. Some macros not supported yet: `$penv{<variable-name>}`, `$vendor{<macro-name>}`
+5. Some macros not supported yet: `$vendor{<macro-name>}`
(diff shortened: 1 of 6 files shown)


Osse:
I've tried this branch and after [two small fixes](https://github.com/lceWolf/cmake-tools.nvim/compare/presets_refactoring...Osse:cmake-tools.nvim:presets_refactoring) it works pretty well for me. Feel free to merge these into yours (or adapt them any way you want (I'm no Lua wizard)). I can make a PR to your branch too if you want.

lceWolf:
Thanks, good catch! I lost these in a refactoring. I amended the fixes to the corresponding commits.

Civitasv:
Thanks for your amazing refactorment.

Civitasv:
Can you resolve the conflicts?

lceWolf:
resolved the conflicts

Civitasv:
Thanks for your contrib, sorry for my late.

PR #257, opened 2024-09-23
feat(terminal): add option to use an alias for terminal runner

lceWolf:
Using an alias allows hiding the lengthy details needed to run the built target.
It also hides the used workaround to catch the command's return code and keeps the command line clean.

Currently this PR does not target windows as I do not have a setup to test a possible implementation. 
I think it should work similar to this PR's approach by using `DOSKEY` tho. 

I am not sure, if the current alias name is the best choice - I am open for suggestions.
Optionally, it could be made configurable by the user.

### DIFF
--- a/lua/cmake-tools/const.lua
+++ b/lua/cmake-tools/const.lua
@@ -112,6 +112,7 @@ local const = {
         start_insert = false, -- If you want to enter terminal with :startinsert upon using :CMakeRun
         focus = false, -- Focus on terminal when cmake task is launched.
         do_not_add_newline = false, -- Do not hit enter on the command inserted when using :CMakeRun, allowing a chance to review or modify the command before hitting enter.
+        use_shell_alias = false, -- Hide the implementation details used to run the built target by using a shell alias
       },
     },
   },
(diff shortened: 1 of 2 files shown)


lceWolf:
should we wait with the PR until a solution has been found to support windows as well?

lceWolf:
I actually missed a crucial part: switching launch targets does not update the alias.
I'll keep this PR on draft until a follow-up is provided.

lceWolf:
Added another function to update the target configuration. If the build dir, launch target or environemnt variables changed, the target configuration is updated.

I tested the new approach for bash and fish shell both with the `use_shell_alias` set to `true` and `false`.
I do not have a windows setup, but comparing the raw command strings did not show any difference.

Still, I would appreciate someone using windows confirming I did not break anything.

Civitasv:
Interesting idea.

Civitasv:
Is this feature stable? Can we merge it?

lceWolf:
Yes, it should be stable. I have been using it without issues since I opened the PR
Windows is not supported tho. If you are fine with this, the PR can be merged

Civitasv:
I think this is fine.

PR #259, opened 2024-10-05
fix: register scratch buffer after changing cwd

Osse:
When starting vim in a project that has a top-level CMakeLists.txt in a subdirectory, then calling :CMakeSelectCwd {subdir} writing to the scratch buffer will fail because it doesn't exist yet. Make sure to register the buffer after changing cwd.


Civitasv:
Good catch!

Issue #261, opened 2024-10-11
build_options containing '--' should be appended at the end


obhi-d:
If you try to pass an option like `'--', '-k', '0'`, it gets 
placed before --target, causing target to be passed to Ninja generator.

Ryex:
also an issue when using it with a VisualStudio generator and you want to pass `{ "--", "/p:UseMultiToolTask=true", "/p:EnforceProcessCountAcrossBuilds=true"}` to get parallel builds

Civitasv:
Should be resolved by #263

Issue #249, opened 2024-08-22
Giving flags to backend generator doesn't work. ( `, "--", "-k 0"`) 


Arniiiii:
**Bug description**
Giving flags to backend generator doesn't work.

**Steps to reproduce**
1. Add `, "--", "-k 0"` or  `, "-- -k 0"` to `:CMakeSettings` -> `return` -> `build_options`

**How it looks like**

https://github.com/user-attachments/assets/44663416-b2ff-4eae-bcdd-fd538b70cf69

**Environment**
- Plugins commit hash: 
 latest at the moment : 4be3c22

PR #263, opened 2024-10-17
append `build_options` last so backend args pass correctly

Ryex:
append the `build_options` args last so that passing `"--", "<some backend arg>"` works correctly

Fix for #249 & #261


Civitasv:
Thanks for your contribution!

Civitasv:
Plz fix the format error.

Issue #262, opened 2024-10-13
CMakeRun error on pwsh, windows


canh25xp:
**Bug description**

CMakeRun give the following error on windows and output is not shown.

```
D:\Test> cmd /C "cd D:\Test\out\Debug\ && D:\Test\out\Debug\Debug/Test.exe" & echo %errorlevel% > C:\Users\Michael\AppData\Local\nvim-tmp-data\cmak
e-tools-tmp\exit_code && del /Q C:\Users\Michael\AppData\Local\nvim-tmp-data\cmake-tools-tmp\.lock 

Id     Name            PSJobTypeName   State         HasMoreData     Location             Command
--     ----            -------------   -----         -----------     --------             -------
5      Job5            BackgroundJob   Running       True            localhost            cmd /C "cd D:\Test\out\D…
Remove-Item: A positional parameter cannot be found that accepts argument 'C:\Users\Michael\AppData\Local\nvim-tmp-data\cmake-tools-tmp\.lock'.
```

It seems like `del /Q` is a cmd command, not a valid pwsh command.

**Steps to reproduce**

minimal config:

```lua
vim.o.shell = "pwsh"
vim.o.shellcmdflag = "-NoLogo -NoProfile -ExecutionPolicy RemoteSigned -Command [Console]::InputEncoding=[Console]::OutputEncoding=[System.Text.Encoding]::UTF8;"
vim.o.shellredir = "-RedirectStandardOutput %s -NoNewWindow -Wait"
vim.o.shellpipe = "2>&1 | Out-File -Encoding UTF8 %s; exit $LastExitCode"
vim.o.shellquote = ""
vim.o.shellxquote = ""

local lazypath = vim.fn.stdpath("data") .. "/lazy/lazy.nvim"
if not (vim.uv or vim.loop).fs_stat(lazypath) then
  local lazyrepo = "https://github.com/folke/lazy.nvim.git"
  local out = vim.fn.system({ "git", "clone", "--filter=blob:none", "--branch=stable", lazyrepo, lazypath })
  if vim.v.shell_error ~= 0 then
    vim.api.nvim_echo({
      { "Failed to clone lazy.nvim:\n", "ErrorMsg" },
      { out, "WarningMsg" },
      { "\nPress any key to exit..." },
    }, true, {})
    vim.fn.getchar()
    os.exit(1)
  end
end
vim.opt.runtimepath:prepend(lazypath)

require("lazy").setup({
  spec = {
    {
[...]

PR #264, opened 2024-10-20
feat: use different remove command for windows power shell (#262)

lceWolf:


Civitasv:
Thanks for your contribution!

canh25xp:
windows has 2 kind of powershell, "pwsh" and "powershell" (the legacy version). So maybe we should check for both?

```lua
local is_power_shell = function()
  return vim.o.shell == "pwsh" or vim.o.shell == "powershell"
end
```

lceWolf:
thanks for the hint. Follow up done in https://github.com/Civitasv/cmake-tools.nvim/pull/267

PR #265, opened 2024-10-20
feat: using correct shell chain symbol for windows powershell (#262)

canh25xp:
`&' is not powershell's chain symbol, but the `;` symbol.

### DIFF
--- a/lua/cmake-tools/terminal.lua
+++ b/lua/cmake-tools/terminal.lua
@@ -630,7 +630,7 @@ function _terminal.run(cmd, env_script, env, args, cwd, opts, on_exit, on_output
   end
 
   -- Send final cmd to terminal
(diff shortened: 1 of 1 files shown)


lua/cmake-tools/terminal.lua
```diff
@@ -630,7 +630,7 @@ function _terminal.run(cmd, env_script, env, args, cwd, opts, on_exit, on_output
   end
 
   -- Send final cmd to terminal
-  local chain_symb = osys.iswin32 and " & " or " ; "
+  local chain_symb = (is_power_shell and " ; ") or (osys.iswin32 and " & " or " ; ")
```
Civitasv, line 633: It should be `is_power_shell()`

PR #272, opened 2024-11-17
fix: incorrect macro expansion for `$(hostSystemName}` on Windows

ayooh:
[Ref](https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_SYSTEM_NAME.html)

### DIFF
--- a/lua/cmake-tools/preset.lua
+++ b/lua/cmake-tools/preset.lua
@@ -1,4 +1,5 @@
 local Path = require("plenary.path")
+local osys = require("cmake-tools.osys")
 
 local Preset = {}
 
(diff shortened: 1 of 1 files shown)


lua/cmake-tools/preset.lua
```diff
@@ -25,7 +25,10 @@ local function expandMacro(self, str)
   if self.generator then
     str = str:gsub("${generator}", self.generator)
   end
-  str = str:gsub("${hostSystemName}", vim.loop.os_uname().sysname)
+  str = str:gsub(
+    "${hostSystemName}",
+    vim.fn.has("win32") == 1 and "Windows" or vim.loop.os_uname().sysname
```
lceWolf, line 30: I think a require for osys would be nice since it would be inline with how other code does this check. Either an "inline require" or one at the top of the file. Maybe something like this
`require("cmake-tools.osys").iswin32 and "Windows" or vim.loop.os_uname().sysname`
or
```
<at the top of the file>
local osys = require("cmake-tools.osys")
...
osys.iswin32 and "Windows" or vim.loop.os_uname().sysname
```

Civitasv:
Thanks!

Civitasv:
Thanks!

Issue #239, opened 2024-06-11
Directory path starting with number causes error


PerMalmberg:
Hi,

Just want to preface this bug report with a huge "thank you" for making this plugin.

**Bug description**
On Windows (shell is pwsh) and on startup, when a project is placed in a folder named `02-breaking` the following error occurs. It appears a path that is begins with numbers is interpreted as as a back reference regex?

```
Error detected while processing C:\Users\perma\AppData\Local\nvim\init.lua:
E65: Illegal back reference
Failed to run `config` for cmake-tools.nvim

...nvim-data/lazy/cmake-tools.nvim/lua/cmake-tools/init.lua:1268: Failed to set autocmd

# stacktrace:
  - cmake-tools.nvim\lua\cmake-tools\init.lua:1268 _in_ **create_regenerate_on_save_autocmd**
  - cmake-tools.nvim\lua\cmake-tools\init.lua:1295 _in_ **register_autocmd**
  - cmake-tools.nvim\lua\cmake-tools\init.lua:54 _in_ **setup**
  - ~/AppData/Local/nvim/lua/plugins/cmake-tools.lua:10 _in_ **config**
  - LazyVim\lua\lazyvim\plugins\extras\lang\cmake.lua:51 _in_ **check**
  - LazyVim\lua\lazyvim\plugins\extras\lang\cmake.lua:55 _in_ **init**
  - ~\AppData\Local\nvim/lua/config/lazy.lua:10
  - ~\AppData\Local\nvim\init.lua:12

```

**Steps to reproduce**

<details><summary>Minimal configuration</summary>
My appologies, I'm just starting out with nvim based on lazynvim so I don't know how to load only `cmake-tools`. My entire config can be found here, if needed: https://github.com/PerMalmberg/nvim-config
</details>

Just open a project (MRP attached) with `cmake-tools` enabled and the error occurs.

I've tried `vim.opt.shellslash = true`, but that makes no difference.

**Expected behavior**
Paths starting with numbers to be handled.

**Environment**

- OS: Windows 11, pwsh
- Plugins commit hash: 09bd614

**Contents in the \*cmake-tools\* buffer**

Sorry, which buffer is this? How do I access it?

[20-breaking.zip](https://github.com/user-attachments/files/15793909/20-breaking.zip)

WayenVan:
Same on windows, my user name start with number so this error occurs everywhere

PR #273, opened 2024-11-18
fix: replace backslashes on Windows for pattern

thisRedH:
On Windows, nvim interprets `\2024` in paths like `C:\foo\2024` as a regex back reference,
causing issues. fixes https://github.com/Civitasv/cmake-tools.nvim/issues/239


Civitasv:
Much Appreciated!

PR #275, opened 2024-11-22
fix: Only switch dir for building with preset

Osse:
config:build_directory_path() is relative to nvim's cwd, not the cwd as given in the config, so cmake --build will fail if config.cwd is a subdirectory of nvim's cwd. On the other hand, when using build presets CMake's cwd *must* be the directory where the preset files are. Fix this by only using config.cwd when building with presets.

### DIFF
--- a/lua/cmake-tools/init.lua
+++ b/lua/cmake-tools/init.lua
@@ -342,15 +342,18 @@ function cmake.build(opt, callback)
   end
 
   local args
+  local cwd
   local presets_exists = config.base_settings.use_preset and Presets.exists(config.cwd)
 
   if presets_exists and config.build_preset then
     args = { "--build", "--preset", config.build_preset } -- preset don't need define build dir.
+    cwd = config.cwd
   else
     args = {
       "--build",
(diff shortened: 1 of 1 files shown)


Civitasv:
Can you give a reproduce method?

Osse:
Sure. Currently on holiday, but I'll be able to in a week or so.

Civitasv:
> Sure. Currently on holiday, but I'll be able to in a week or so.

Ok, enjoy your holiday 😄

Osse:
I have tested more, and I found that my change fixes _something_ but I'm not quite sure what yet :smile: Will need to do some more testing. Please don't merge yet.

Also, formatting fails but the code that fails is not mine, so I'm not sure what you prefer I do about that.

Civitasv:
> I have tested more, and I found that my change fixes _something_ but I'm not quite sure what yet 😄 Will need to do some more testing. Please don't merge yet.

ok, I'll wait.



> Also, formatting fails but the code that fails is not mine, so I'm not sure what you prefer I do about that.

You can also fix that if not too much. :smile:

Osse:
I have decided to close this PR and rather make an issue of my findings.

PR #276, opened 2024-11-22
Don't make binaryDirExpanded relative

Osse:
It's made relative to nvim's cwd, not config.cwd and furthermore there is no point in making it relative in the first place.


Civitasv:
@lceWolf Can you review this?

lceWolf:
I think it is fine to omit making it relative. If an absolute path is given, it should be used as is. Only if the binaryDir is given as relative it should be treated relative to the source dir. Transforming the absolute path into a relative one and concatenating this with `config.cwd` is wrong indeed.

lceWolf:
Thanks fixing that - good catch :)

PR #277, opened 2024-11-22
fix: Change defaults to match README

Osse:
cb7dc21a6da58f improved the defaults for overseer, but omitted to set actually set on_new_task to the value stated in the README.


Civitasv:
Thanks!

PR #283, opened 2024-12-05
Fix CMakeRun error in windows, Caused by slash.

qiyuewuyi2333:
Fix the path slash error on windows when use CMakeRun.

Issue #282, opened 2024-12-04
CMakeRun doesn't work in my windows11 laptop


qiyuewuyi2333:
**Bug description**

When I use CMakeRun command in nvim to quick run my Cpp project on windows, I got the output as bellow image:
![image](https://github.com/user-attachments/assets/7b8dfa18-e169-48c1-966a-c53e7d030227)

The command runs good in my linux vm. But It seems like the command is not compatible well for windows path.

But weirdly, I can run '.\bin/******.exe' on the Clang_Debug dir path. I don't clear that why windows doesn't recoginize it.

**Steps to reproduce**

<details><summary>Minimal configuration</summary>
</details>
Just a normal config for windows.
I don't think it could be cause by other plugins.
here is my configuations:

[github pos](https://github.com/qiyuewuyi2333/nvim/blob/master/lua/huan/plugins/cmake-tools.lua)

**Environment**
OS: windows11 24H2
cmake-tools version: 
![image](https://github.com/user-attachments/assets/2c9faf29-deec-4866-a089-5c58f690e030)

qiyuewuyi2333:
Explain for the first image:
'.\bin' is not a program executor or command

qiyuewuyi2333:
I think it may caused by cmd. Because It's ok to run it in pwsh.

qiyuewuyi2333:
the .\bin/******.exe should be modify to .\bin\******.exe for find the execute

qiyuewuyi2333:
I made PR for fix this error. 

#283

Issue #269, opened 2024-10-22
CMakeRun command broken by .lock file


Alan-John-Byrne:
**Bug description**
Running the CMakeRun command after building the executable in a powershell environment will result in an error within the terminal that's opened, which is supposed to then output the result of the program executed.

**Steps to reproduce**

NOTE: THIS ISSUE WAS FOUND IN A WINDOWS POWERSHELL ENVIRONMENT.

1. Make a CMakeLIsts.txt and populate it with the desired cpp project specs.
2. Run the CMakeGenerate command to generate the build files. (After first setting the generator and build directory in CMakeSettings).
3. Run the CMakeBuild command to produce an executable from your project.
4. Run the CMakeRun, which is intended for executing that produced executable.

**Expected behavior**
I expected the output of my program from main.cpp to appear. So for example, "Hello World!".

**Screenshots**
![image](https://github.com/user-attachments/assets/0b4dc90a-6a0b-4cea-97e7-cfb7dcfb3b65)

**Environment**

- OS: Windows 11
- Desktop environment: Windows 11
- Plugins commit hash: (PLEASE SEE THE IMAGE BELOW)
![image](https://github.com/user-attachments/assets/43683650-5ec9-4ff1-83f9-a433e112d293)

**Contents in the \*cmake-tools\* buffer**

I'm sorry I'm uncertain what this section means, however the error we are worried about is below:

***Remove-Item: A positional parameter cannot be found that accepts argument 'C:\Users\alanj\AppData\Local\nvim-data\cmake-tools-tmp\.lock'.***

The above error is outputted in the terminal that's opened upon running the 'CMakeRun' command. When we should be seeing the output of the program.

pcorbineau:
Got quite a similar issue on last master on cmd
![image](https://github.com/user-attachments/assets/3f6cde92-ed87-44d5-a527-feaf47830108)

path are not Windows friendly ...

qiyuewuyi2333:
Me too. It hinds me the .\bin is not an program or commands

CruizK:
For this specific error you need to specify `vim.opt.shell = 'pwsh'` so it will use `Remove-Item` instead of `del`

PR #285, opened 2025-01-04
fix(windows): Fixed cmake run command using forward slashes instead of backslashes

CruizK:
Unsure of the status of #283, but this is the first change in that PR without all the added code (which I think is related to auto-session, but unsure?). This should be the minimum to get windows working again, as its currently broken on master.

This should solve: #282 & #269

### DIFF
--- a/lua/cmake-tools/config.lua
+++ b/lua/cmake-tools/config.lua
@@ -258,6 +258,9 @@ end
 
 function Config:get_launch_target_from_info(target_info)
   local target_path = target_info["artifacts"][1]["path"]
+  if require('cmake-tools.osys').iswin32 then
+    target_path = target_path:gsub("/", "\\")
+  end
   target_path = Path:new(target_path)
   if not target_path:is_absolute() then
     -- then it is a relative path, based on build directory
--- a/lua/cmake-tools/config.lua
+++ b/lua/cmake-tools/config.lua
@@ -258,7 +258,7 @@ end
 
 function Config:get_launch_target_from_info(target_info)
   local target_path = target_info["artifacts"][1]["path"]
-  if require('cmake-tools.osys').iswin32 then
+  if require("cmake-tools.osys").iswin32 then
     target_path = target_path:gsub("/", "\\")
   end
   target_path = Path:new(target_path)
--- a/lua/cmake-tools/init.lua
+++ b/lua/cmake-tools/init.lua
@@ -1403,7 +1403,7 @@ function cmake.create_regenerate_on_save_autocmd()
 
   local cwd = config.cwd
   if require("cmake-tools.osys").iswin32 then
-      cwd = config.cwd:gsub("\\", "/")
+    cwd = config.cwd:gsub("\\", "/")
   end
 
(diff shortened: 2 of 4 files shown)


lceWolf:
Wouldn't all runners suffer from this problem? I think the `cmd` has to be patched before passing it to a runner, right?

CruizK:
> Wouldn't all runners suffer from this problem? I think the `cmd` has to be patched before passing it to a runner, right?

Good point, Strangely enough I was not able to actually get this to error out on anything that wasn't terminal, but to be complete it should be done to all runners.

 I moved it so its patched during `utils.run` instead, right before the specific runner would get called

lceWolf:
Thanks for the changes!

Weird it is only happening for the terminal runner... Anyways, better to be safe for all runners.
Did you dig deeper by chance and identified the root cause? I wonder were the wrong path comes from 🤔 

IMO, the changes in this MR are fine for getting the issue fixed, but the formatter is not happy. Can you fix this please?

``` 
Diff in ./lua/cmake-tools/utils.lua:
189 189  |   vim.cmd("silent exec " .. '"wall"')
190 190  | 
191 191  |   if osys.iswin32 then
192      |-    cmd = cmd:gsub('/', '\\')
    192  |+    cmd = cmd:gsub("/", "\\")
193 193  |   end
194 194  | 
195 195  |   local ntfy = notification:new("runner")
Diff in ./lua/cmake-tools/init.lua:
14031403 | 
14041404 |   local cwd = config.cwd
14051405 |   if require("cmake-tools.osys").iswin32 then
1406     |-      cwd = config.cwd:gsub("\\", "/")
    1406 |+    cwd = config.cwd:gsub("\\", "/")
14071407 |   end
14081408 | 
14091409 |   local presets_exists = config.base_settings.use_preset and Presets.exists(config.cwd)
```

CruizK:
So I was able to actually find the root cause of the bug (Also fixed linting). CMake gives the path with `/` when you query the path info.
For comparisons without the fix it spits out
```
cmd /C "cd C:\path\to\build\ && .\Project/Project.exe"
```
with the utils fix it will give
```
cmd /C "cd C:\path\to\build\ && .\Project\Project.exe"
```
which works, but does not seem what these 2 paths are really meant to mean. But if we patch it right after we query cmake we get
```
cmd /C "cd C:\path\to\build\Project\ && .\Project.exe"
```
Which I would think is the way it should work.

lua/cmake-tools/config.lua
```diff
@@ -258,6 +258,9 @@ end
 
 function Config:get_launch_target_from_info(target_info)
   local target_path = target_info["artifacts"][1]["path"]
+  if require('cmake-tools.osys').iswin32 then
```
lceWolf, line 261:
```suggestion
  if require("cmake-tools.osys").iswin32 then
```
stylua requirement

lceWolf (changes requested):

lua/cmake-tools/config.lua
CruizK: Oops, fail. fixed.

lceWolf:
Thanks for taking care of the issue!

PR #286, opened 2025-01-12
fix(windows): Fix powershell legacy and operator `&&`

canh25xp:
The windows powershell legacy ( version 5.1 that come default with windows) does not have the `&&` operator which cause `CMakeRun` command error:

```
C:\Users\Michael\projects\cpp\DSA> cmd /C "cd C:\Users\Michael\projects\cpp\DSA\./build\Pro\Bacteria\ && .\Bacteria.exe"; echo $LASTEXITCODE > C:\Users\Michael\AppData\Local\nvim-data
\cmake-tools-tmp\exit_code && Remove-Item C:\Users\Michael\AppData\Local\nvim-data\cmake-tools-tmp\.lock 
At line:1 char:176
+ ... \Michael\AppData\Local\nvim-data\cmake-tools-tmp\exit_code && Remove- ...
+                                                                ~~
The token '&&' is not a valid statement separator in this version.
    + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : InvalidEndOfLine
```

So I suggest using the `-and` operator since both the legacy and the newer version of powershell support this.

### DIFF
--- a/lua/cmake-tools/terminal.lua
+++ b/lua/cmake-tools/terminal.lua
@@ -532,6 +532,7 @@ local get_command_handling_on_exit = function()
 
   local exit_op = "$?"
   local escape_rm = " \\rm -f "
+  local and_op = " &&"
 
   if is_fish_shell() then
     exit_op = "$status"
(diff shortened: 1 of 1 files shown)


lceWolf:
Thank you for your contribution!

PR #287, opened 2025-01-14
fix: pass config argument to build process for multiconfig generators

sebu06:
Multiconfig generators like ninja need the build type passed at the build stage and not the configure stage.


lceWolf:
Thank you! :)

PR #295, opened 2025-02-05
Better handling of build preset targets

arntanguy:
In the current implementation, the `targets` field of `CMakePresets.json` is ignored. This leads to inconsistencies between what `cmake-tools.nvim` does and displays and what the user expects to happen.

Here is an example `buildPresets` showcasing the 4 possible cases:
1. One target specified as a string
2. One target specified as an array with a single string
3. Several targets specified
4. No target specified

```json
  "buildPresets": [
    {
      "name": "build-one-string",
      "configurePreset": "release",
      "configuration": "Release",
      "targets": "target1"
    },
    {
      "name": "build-one-array",
      "configurePreset": "release",
      "configuration": "Release",
      "targets": ["target1"]
    },
    {
      "name": "build-two",
      "configurePreset": "release",
      "configuration": "Release",
      "targets": ["target1", "target2"]
    }
    {
      "name": "build-notarget",
      "configurePreset": "release",
      "configuration": "Release",
    }
  ]

```

One expects that : 
- when calling `:CMakeSelectBuildPreset` the internal `build_target` to be updated. This ensures that calling `get_build_target()` returns the target specified in the preset (for lualine for instance).
- when calling `:CMakeBuild` the command executed to be `cmake --preset <preset_name>` without additional `--target` arguments

The current behaviour priour to this PR is to:
- Disregard the `targets` field of the `CMakePresets.json`
- Run `cmake --preset <preset_name> --target <whatever the previously specified target was>`

For a more concrete use-case, consider embeded development with a `build-preset` preset with a `build-target` target and and a `flash-preset` preset with a `flash-target` target that flashes the program onto the device. 

```json
  "buildPresets": [
    {
      "name": "build-preset",
      "configurePreset": "release",
      "configuration": "Release",
      "targets": "build-target"
    },
    {
      "name": "flash-preset",
      "configurePreset": "release",
      "configuration": "Release",
      "targets": ["flash-target"]
    },
```

Currently, the following would happen:
- `:CMakeSelectBuildPreset build-preset`
  - the `build-preset` is indeed selected
  - `build_target`: none  => we expected `build-target` instead
- `:CMakeBuild` will prompt for a target, we select `build-target`, and `cmake --preset build-preset --target build-target` is run
- `:CMakeSelectBuildPreset flash-preset`
  - the `flash-preset` is indeed selected 
  - `build_target`: `build-target` => we expected `flash-target`
  - `:CMakeBuild` runs `cmake --preset flash-preset --target build-target` => we expected to run the `flash-target`

You get the gist. This PR fixes this behaviour in the following way:
- When a new build preset is selected, the `build_target` is changed to what's specified in that preset with one caveat: multiple targets `targets: ["target1", "target2"]`. Since this case is not handled anywhere else in the code, I simply put the `build_target = nil`
- The user only gets asked for which target to select if none where defined in the selected preset
- The build command is always `cmake --preset <selected_preset>` if at least one target is specified in the preset, otherwise it is `cmake --preset <selected_preset> --target <user selected target>`

It would probably be better to have a proper way of handle the multiple-targets case. However this seemed too big of a change, for little benefit for me to tackle right now.

### DIFF
--- a/lua/cmake-tools/init.lua
+++ b/lua/cmake-tools/init.lua
@@ -331,7 +331,9 @@ function cmake.build(opt, callback)
     end)
   end
 
-  if opt.target == nil and config.build_target == nil then
+  local presets = Presets:parse(config.cwd)
+  local build_preset = presets:get_build_preset(config.build_preset)
+  if (not config.build_preset or config.build_preset and not build_preset.targets) and opt.target == nil and config.build_target == nil then
     return cmake.select_build_target(true, function(result)
       if result:is_ok() then
         cmake.build(opt, callback)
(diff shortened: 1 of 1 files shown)


lua/cmake-tools/init.lua
```diff
@@ -825,8 +830,24 @@ function cmake.select_build_preset(callback)
...
+        if build_preset.targets then
+          if type(build_preset.targets) == string then
+            config.build_target = build_preset.targets
+          else -- build preset is an array
+            if #(build_preset.targets) == 1 then
+              config.build_target = build_preset.targets[1]
+            else
+              -- XXX: Array targets are not supported in other code paths
+              -- As a workaround, unset the current build_target if any (it is not used when using build presets)
+              config.build_target = nil
```
lceWolf, line 843: as we do not have a handling for multiple build targets currently, you could either keep the currently selected one (if it is part of the array) or always select the first one from the list. The user can still change the target, if needed.
Resetting the selected target is not desirable, I think

I think a refactoring to support multiple targets might be helpful. But this is beyond the scope of this MR for sure.

lceWolf:
I appreciate the PR and think it is a step in the right direction for sure. Very nice write-up of the motivation and your approach!
Thanks for taking care of that topic.

PR #297, opened 2025-02-12
Add support for CMAKE_BUILD_TYPE presets using "type"/"value" form.

acalhoon:
With the current implementation, if CMakePresets.json sets CMAKE_BUILD_TYPE as a JSON object with a "type" and "value", then it causes the build/configuration fail -- example below.

```json
"CMAKE_BUILD_TYPE": {
  "type": "STRING",
  "value": "Debug"
}
```

This fix looks for a CMAKE_BUILD_TYPE with a "table" type and extracts the "value" element, otherwise it falls back to the original behavior.

### DIFF
--- a/lua/cmake-tools/preset.lua
+++ b/lua/cmake-tools/preset.lua
@@ -4,8 +4,8 @@ local osys = require("cmake-tools.osys")
 local Preset = {}
 
 local function expandMacro(self, str)
-  if type(str) == "table" and str["value"] ~= nil then
-    str = str["value"]
+  if type(str) == "table" and str.value ~= nil then
+    str = str.value
   end
 
   if type(str) ~= "string" then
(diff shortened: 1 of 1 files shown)


lceWolf:
Thanks for your contribution. I think handling cacheVariables in the type and value form should be done when parsing the preset itself instead. That way this form can be supported for each cacheVariable without special handling by the consumer of that variable.

Mind adapting the cacheVariable parsing accordingly?

acalhoon:
I like the suggestion and have pushed a change that I think will work but still need to test it. Marked as draft accordingly.

lua/cmake-tools/preset.lua
```diff
@@ -4,6 +4,10 @@ local osys = require("cmake-tools.osys")
 local Preset = {}
 
 local function expandMacro(self, str)
+  if type(str) == "table" and str["value"] ~= nil then
```
lceWolf, line 7: can you please use `str.value` for both cases?

acalhoon: Addressed.

lceWolf:
Thanks for your contribution!

PR #299, opened 2025-02-17
fix: Fix error message when Overseer is running a job

Osse:
`command` is not a field on `overseer.Task` leading to this error
originating from check_active_job_and_notify():

    "Error executing Lua callback: .../nvim/lazy/cmake-tools.nvim/lua/cmake-tools/overseer.lua:55: attempt to concatenate field 'command' (a nil value)"

But `cmd` is a valid field. Though if the user gives the task a custom
name then that's probably what they want to see.


lceWolf:
Thanks!

PR #301, opened 2025-02-18
Added option to override config with local file.

sebu06:
Since I often use cmake-tools on different machines, I would like to have an option to store my settings in the repo rather than in the session store, that is only locally available. This commit adds the option to add a file ".cmake-tools.lua" to the repo and override the cmake-tools settings in it.

### DIFF
--- a/lua/cmake-tools/session.lua
+++ b/lua/cmake-tools/session.lua
@@ -1,6 +1,22 @@
 local osys = require("cmake-tools.osys")
 local utils = require("cmake-tools.utils")
 
+local function find_file_in_dir_or_parent(dir, fname)
+  local path = dir .. "/" .. fname
+  local file = io.open(path, "r")
+  if file then
+    file:close()
+    return path
+  end
+
+  local parentDir = dir:match("(.+)/[^/]+$")
+  if not parentDir then
+    return nil
+  end
+
+  return find_file_in_dir_or_parent(parentDir, fname)
+end
+
 local session = {
   dir = {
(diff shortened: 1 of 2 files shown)


lua/cmake-tools/session.lua
```diff
@@ -1,6 +1,22 @@
 local osys = require("cmake-tools.osys")
 local utils = require("cmake-tools.utils")
 
+local function findFileInDirOrParent(dir, fname)
```
Civitasv, line 4: Change it to `find_file_in_dir_or_parent`.

Civitasv: And I think it's better to just search `.cmake-tools.lua` from current working directory. Current approach may lead to recursion bug. Sometimes we better restrict our power.

Civitasv:
Currently, it seems will still create cache file even .cmake-tools.lua file exists, which I don't think quite good.

lua/cmake-tools/session.lua
sebu06: > Change it to `find_file_in_dir_or_parent`.

done

> And I think it's better to just search `.cmake-tools.lua` from current working directory. Current approach may lead to recursion bug. Sometimes we better restrict our power.

I find it rather useful, since it allows me to set good presets for all my repos in the parent folder and fine tune them in the individual ones if needed...

lceWolf: I too think this search should be limited (similar how `CMakeLists.txt` is found)

lceWolf:
@sebu06 sorry for the late reply. Mind resolving the merge conflicts?

PR #302, opened 2025-02-19
Make compile_commands.json symlinking "silent"

acalhoon:
Prior to this change, the user is forced to acknowledge the creation of the symlink.

Now, the symlink happens "silently" in the background without extra user steps.


lceWolf:
Personally, I actually like to see what is going on. Hence I am not too big of a fan to supress the output tbh.
@Civitasv WDYT?

Civitasv:
Maybe make it an option? I also prefer to see what is going on.

lceWolf:
@acalhoon any news on this topic? Do you want to persue it further?

PR #304, opened 2025-02-28
fix: Indexing a nil value when no target launch settings are configured

omgitsmoe:
When running a specific target (e.g. `CMakeQuickRun`) there might not be a `target_settings` config available. Previously the `target_settings` of that target were always accessed assuming it was there.
Now each access is checked and an empty table is used as a fallback.

Feel free to make any changes or suggestions!


lceWolf:
Can you address the formatting issues please?
LGTM otherwise. Thanks!

omgitsmoe:
Hope the formatting is fine now. Thank you for the quick response!

PR #308, opened 2025-03-08
fix(terminal): exit from insert mode if terminal is not focused or start_insert is false

acrsantos:


lua/cmake-tools/terminal.lua
```diff
@@ -209,6 +211,7 @@ function _terminal.send_data_to_terminal(buffer_idx, cmd, opts)
       -- cursor out of the cmake buffer, as it can be annoying
       if opts and not opts.focus then
         vim.cmd("wincmd p") -- Goes back to previous window: Equivalent to [[ CTRL-W p ]]
+        vim.cmd("stopinsert!")
```
lceWolf, line 214: wouldn't that try to leave insert mode after switching windows? I.e. isn't it trying to stop it in the wrong window?

lua/cmake-tools/terminal.lua
```diff
@@ -198,6 +198,8 @@ function _terminal.send_data_to_terminal(buffer_idx, cmd, opts)
     vim.api.nvim_set_current_win(opts.win_id)
     if opts.start_insert then
       vim.cmd("startinsert")
+    else
+      vim.cmd("stopinsert!")
```
lceWolf, line 202: Do we need this if we stop insert mode when leaving the terminal?

acrsantos:
Whoops, you're right, this PR is unnecessary. I forgot I have autocmd that switch to insert mode when entering a terminal. It seems that it retains in insert mode after switching to previous window, causing me to be in insert mode on the buffer I am currently in. I should've looked into it more, my bad. Thanks!

PR #309, opened 2025-04-05
Feature | Adds command to open the cmake cache file

BenLeadbetter:
## The Feature

Adds a new command to open the cmake cache file. CLion has this functionality which I always found super helpful.

## Considerations

In this initial implementation the file is opened simply with a vim `edit <cache file>` command. Being a vim classicist this suites my workflow, but I can see that other commands (for example CMakeOpenRunner) open a new split window. Would this be preferable?


BenLeadbetter:
> What is your way of working with that file?

I'm often editing the cache file to switch on and off the various configuration options a project might expose during development. After editing cmake is smart enough to detect the change and regenerate the next time a build is kicked off.

PR #314, opened 2025-04-25
Exposed config.cwd for the user

ReKylee:
It really annoyed me how when opening a single cpp file it couldn't find the base CMakeLists.txt, so I did some changes:

- removed deprecated nvim commands and replaced with their new counterpart
- exposed cwd to be accessible from opts

I should mention this still **doesn't fully work**, it seems like the path gets overridden at some point and idk why. 
Also should mention this is my first time trying to maintain anything lua related lmao.
If anyone is willing to help on this I'd appreciate it a lot, I find it so disruptive that I can't just open a file and it find out where the CMakeLists.txt file is, sets up the build directory correctly, and generally doesn't cooperate when opening a file that doesn't immediately set cwd to be the root directory.

### DIFF
--- a/lua/cmake-tools/config.lua
+++ b/lua/cmake-tools/config.lua
@@ -29,7 +29,7 @@ local Config = {
   executor = nil,
   runner = nil,
   env_script = " ",
-  cwd = vim.loop.cwd(),
+  cwd = vim.fn.getcwd(),
 }
 
 function Config:new(const)
--- a/lua/cmake-tools/config.lua
+++ b/lua/cmake-tools/config.lua
@@ -29,7 +29,7 @@ local Config = {
   executor = nil,
   runner = nil,
   env_script = " ",
-  cwd = vim.fn.getcwd(),
+  cwd = nil,
 }
 
 function Config:new(const)
--- a/lua/cmake-tools/const.lua
+++ b/lua/cmake-tools/const.lua
@@ -1,5 +1,6 @@
 local osys = require("cmake-tools.osys")
 local const = {
+  cwd = vim.fn.getcwd(),
   cmake_command = "cmake", -- this is used to specify cmake command path
   ctest_command = "ctest", -- this is used to specify ctest command path
   cmake_use_preset = true, -- when `false`, this is used to define if the `--preset` option should be use on cmake commands
--- a/lua/cmake-tools/init.lua
+++ b/lua/cmake-tools/init.lua
@@ -24,6 +24,9 @@ local cmake = {}
 
 --- Setup cmake-tools
 function cmake.setup(values)
+  if values and type(values.cwd) == "function" then
+    values.cwd = values.cwd()
+  end
   const = vim.tbl_deep_extend("force", const, values)
   const.cmake_executor.opts = vim.tbl_deep_extend(
     "force",
@@ -35,11 +38,12 @@ function cmake.setup(values)
     const.cmake_runner.default_opts[const.cmake_runner.name],
     const.cmake_runner.opts or {}
   )
-
   require("cmake-tools.notification").setup(const.cmake_notifications)
 
   config = Config:new(const)
 
+  config.cwd = const.cwd
+  print("Using cwd:", config.cwd)
   -- auto reload previous session
   local old_config = _session.load()
   _session.update(config, old_config)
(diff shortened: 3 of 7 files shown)


Civitasv:
Have you tried CMakeSelectCwd?

ReKylee:
It doesn't matter if I use SelectCwd or not unfort.
I can't seem to use it from my config, and even when I manually set it, the build dir doesn't get updated and it's trying to set it inside my Source folder.
Note I tried both this config and manually with a command.

```
return {
  "Civitasv/cmake-tools.nvim",
  version = false,
  opts = function(_, opts)
    local root = require("lazyvim.util.root").get({ normalized = true })
    require("cmake-tools.init").select_cwd(root)
    return opts
  end,
}
```

Auto generated Source/out right after I did CMakeSelectCwd:
{071C7CE0-7B69-47B6-8F6E-DD5B29508014}

Generally yeah I also agree I'm pretty dumb for forgetting CMakeSelectCwd exists but I've noticed that if it's not selected right on startup it gets overridden with vim.loop.cwd() which is both deprecated and doesn't actually select the cwd that is reported by vim, but selects the cwd that is reported by the process itself.

Civitasv:
> I can't seem to use it from my config, and even when I manually set it, the build dir doesn't get updated and it's trying to set it inside my Source folder.

You should use CMakeSelectBuildDir to change build directory. See https://github.com/Civitasv/cmake-tools.nvim/blob/master/docs/all_commands.md.

ReKylee:
```
return {
  "Civitasv/cmake-tools.nvim",
  version = false,
  opts = function(_, opts)
    opts.cmake_generate_options = {
      "-DCMAKE_EXPORT_COMPILE_COMMANDS=1",
      "-G",
      "Ninja",
      "-DCMAKE_C_COMPILER=clang",
      "-DCMAKE_CXX_COMPILER=clang++ ",
    }
    opts.cmake_dap_configuration = {
      name = "cpp",
      type = "codelldb",
      request = "launch",
      stopOnEntry = false,
      runInTerminal = true,
      console = "integratedTerminal",
    }
    local root = require("lazyvim.util.root").get({ normalize = true })
    print(root)
    require("cmake-tools").select_cwd(root)
    require("cmake-tools").select_build_dir(root)
  end,
}
```

And yet:
{421E6DE7-1675-4A0D-84D2-E32A057D4D81}

I've also tried:
```
opts = function(_, opts)
    opts.cmake_generate_options = {
      "-DCMAKE_EXPORT_COMPILE_COMMANDS=1",
      "-G",
      "Ninja",
      "-DCMAKE_C_COMPILER=clang",
      "-DCMAKE_CXX_COMPILER=clang++ ",
    }
    opts.cmake_dap_configuration = {
      name = "cpp",
      type = "codelldb",
      request = "launch",
      stopOnEntry = false,
      runInTerminal = true,
      console = "integratedTerminal",
    }
    LazyVim.on_load("cmake-tools.nvim", function()
      local root = require("lazyvim.util.root").get({ normalize = true })
      print(root)
      require("cmake-tools").select_cwd(root)
      require("cmake-tools").select_build_dir(root)
    end)
  end,
```
But nope.

Also, places where we don't use config.cwd at all:
{3B1CABA4-1B7A-4EA1-B572-EB5A1F883474}

ReKylee:
Update:
Apparently, I had to pass {args = root} instead of just root.
Still, here's what I get and also it's not consistent:
```
local root = require("lazyvim.util.root").get({ normalize = true })
    print(root)
    require("cmake-tools").select_cwd({ args = root })
    require("cmake-tools").select_build_dir({ args = root })
```
The error:
```
Failed to run `config` for cmake-tools.nvim

vim/_editor.lua:447: nvim_exec2()[1]..BufReadPost Autocommands for "*"..FileType Autocommands for "cpp"..script nvim_exec2() called at FileType Autocommands for "cpp":0, line 1: Vim(wall):E141: No file name for buffer 8

# stacktrace:
  - vim\_editor.lua:447 _in_ **cmd**
  - cmake-tools.nvim\lua\cmake-tools\utils.lua:229 _in_ **generate**
  - cmake-tools.nvim\lua\cmake-tools\init.lua:1366 _in_ **select_cwd**
  - P:/.config/nvim/lua/plugins/coding/cmake.lua:22 _in_ **values**
  - P:/.config/nvim/lua/plugins/coding/clangd.lua:68
```

Civitasv:
> This only works when you open a file through nvim, and not when you open it externally:

Yes, currently, that's how you change cwd and build_dir. But if you change once, next it will remember the right "workspace" for this file. Does this address your needs?

Alternatively, we could implement a prompt when user open an external file but cannot find CMakeLists.txt. The prompt could ask:

"No CMakeLists.txt found here. A parent directory has one at [path]. Set workspace to this location?"

ReKylee:
I think I'd at minimum prefer the popup. I honestly just wanna be able to open a cpp file from my explorer, have it find the CMakeLists.txt and compile ^-^;

Issue #296, opened 2025-02-09
Option to pass extra arguments to ctest


kalebealvs:
As far as I could see, the current behavior doesn't allow for extra arguments to be given to ctest. This is a bit limiting. Personally, I find it very annoying to have to search for output logs to get more in-depth test reports, as well as have some logger tests being executed when stopping on previous errors would be desired. All that can be fixed with a couple extra arguments.

A simple solution would be to have an extra field on the setup config object perhaps called ctest_extra_args. 

I thought about having it in the executor opts, but that doesn't make much sense since reading the documentation this seems to be intended to be passed to the executor plugin only.
Alternatively, running the ctest module with some environment variable would work as well, but this looks less flexible and unless it is targeted at the ctest module, would possibly contaminate other environments unnecessarily.

I created a simple implementation for that proposal, which for now is working fine for me, but I think this is generic enough so others might benefit from it as well. Also having it upstreaming is much better than having to sync my forked version.
It's been nearly a decade since I seriously programmed in Lua, so I'm not up-to-date with best practices. Any input is highly appreciated here :D.

This is the [link](https://github.com/kalebealvs/cmake-tools.nvim/commit/13b9f37a611c79fd95c8509fa81d57343616017a) for the changes I made.

lceWolf:
You can pass additional arguments to `CMakeRunTest` like `CMakeRunTest -V --debug`. That said, I kinda like the idea of having additional arguments which gets applied for every invocation.
@Civitasv WDYT?

Your changes look clean and concise. And you even got some tests - sweet!

bbnolauv:
I'm also looking forward to this.

lceWolf:
@kalebealvs mind creating a PR for this?
What I thought of in the meantime is the following: Instead of always appending the extra args "invisibly", we could add them to the prompt so that the user can still "opt out" in some cases by deleting an argument manually.
WDYT?

PR #315, opened 2025-05-02
Option to pass extra arguments to ctest

kalebealvs:
This patch enables the plugin to pass extra args to ctest when running tests.

Resolves: #296

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ local osys = require("cmake-tools.osys")
 require("cmake-tools").setup {
   cmake_command = "cmake", -- this is used to specify cmake command path
   ctest_command = "ctest", -- this is used to specify ctest command path
+  ctest_extra_args = {}, -- this is used to specify extra args for ctest
   cmake_use_preset = true,
   cmake_regenerate_on_save = true, -- auto generate when save CMakeLists.txt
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -28,6 +28,7 @@ local osys = require("cmake-tools.osys")
 require("cmake-tools").setup {
   cmake_command = "cmake", -- this is used to specify cmake command path
   ctest_command = "ctest", -- this is used to specify ctest command path
+  ctest_extra_args = {}, -- this is used to specify extra args for ctest
```
Civitasv, line 31: Maybe we can put this variable into `Config.base_settings`? So we can edit it easily.

kalebealvs:
> Sorry for my late. It's a great option, I personally think putting it into `cmake_settings` will be better. What do you think?

I'm so sorry. I completely forgot about this since I was using my fork and didn't pay much attention to the notifications.
Do you mean creating a new object called cmake_settings and putting ctest_extra_args under it?
That being the case, isn't this root object kinda like "cmake settings" already, since there are other things here also tied to cmake settings, like regenerate_on_save, generate_options, build_directory, among others?

README.md
kalebealvs: I like the idea. Done.

lceWolf (commented):
Sorry for my late reply. I think this PR is a move in the right direction, I only got some minor comments.

Unfortunately, there are merge conflicts now which need resolving.

Issue #243, opened 2024-06-21
Copy compile_commands to a user specified directory after CMake Configure


SineStriker:
Same as `cmake.copyCompileCommands` in VSCode cmake tools.

kamou:
if you are using clangd, you may want to create the .clangd file at the root of your project with the following content:
```
CompileFlags:
  CompilationDatabase: out/Debug
```

PR #316, opened 2025-05-12
feat: supports copy compile commands and specify target directory

Elivicti:
Replaced `cmake_soft_link_compile_commands` and `cmake_compile_commands_from_lsp` with `cmake_compile_commands_options`, where you can specify action and target path:
```lua
cmake_compile_commands_options = {
  action = "soft_link", -- available options: soft_link, copy, lsp, none
                        -- soft_link: this will automatically make a soft link from compile commands file to target
                        -- copy:      this will automatically copy compile commands file to target
                        -- lsp:       this will automatically set compile commands file location using lsp
                        -- none:      this will make this option ignored
  target = vim.loop.cwd() -- path to directory, this is used only if action == "soft_link" or action == "copy"
},
```
With `action = "soft_link"`, the behavior is the same as
```lua
cmake_soft_link_compile_commands = true
```
With `action = "lsp"`, the behavior is the same as
```lua
cmake_soft_link_compile_commands = false,
cmake_compile_commands_from_lsp = true
```
With `action = "none"`, the behavior is the same as
```lua
cmake_soft_link_compile_commands = false,
cmake_compile_commands_from_lsp = false
```
Default behavior remains unchanged. The `copy` action is my personal preference, and is also requested by #243 .

`target` is path to directory where the created symlink file or copied compile_commands.json is stored. Previously it is hardcoded to `vim.loop.cwd()`, now it's configurable.

I'm really bad at naming things, so if you don't like the naming or any other thing, feel free to give suggestions or just close PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,8 +42,14 @@ require("cmake-tools").setup {
     end
     return "out/${variant:buildType}"
   end, -- this is used to specify generate directory for cmake, allows macro expansion, can be a string or a function returning the string, relative to cwd.
-  cmake_soft_link_compile_commands = true, -- this will automatically make a soft link from compile commands file to project root dir
-  cmake_compile_commands_from_lsp = false, -- this will automatically set compile commands file location using lsp, to use it, please set `cmake_soft_link_compile_commands` to false
+  cmake_compile_commands_options = {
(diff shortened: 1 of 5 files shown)


Civitasv:
Thanks for your contribution! This is a nice feature.

PR #317, opened 2025-05-14
fix: only append when scratch.buffer is not nil

ChillyWall:
The new feature cmake_use_scratch_buffer lets user can decide whether to use the scratch buffer. @Civitasv ensures that the scratch is created only when cmake_use_scratch_buffer is true. But no matter whether this scratch buffer is created, scratch.append would be called when new operation is performed.
Thus, I add a check to ensure that scratch.append would be called only when scratch buffer exists.


Civitasv:
Thanks for your fix.

PR #332, opened 2025-07-26
Fix build presets with multiple targets

bornabesic:
If a build preset from CMakePresets.json contained multiple targets, `BuildPreset:get_build_target` would concatenate target names into a single string which would be used as a single command line argument. The build system, such as Ninja, would then complain that the specified target is unknown.

This change modifies `BuildPreset:get_build_target` to return a table of targets instead, which is then unpacked after CMake's `--target` flag. That way, each target is a separate command line argument and the build system executes a proper multi-target build.


lceWolf:
lgtm. Thanks for your contribution

Issue #333, opened 2025-08-02
CMake build presets are required when using configure presets


bailwillharr:
**Bug description**

It is very common for projects to have a CMakePresets.json file that has configure presets but no build presets.

When attempting to run :CMakeBuild on a project with a CMakePresets.json file like this, cmake-tools correctly prompts for the configure preset, but then it prompts for a build preset from an empty list. This makes it impossible for the build to actually run as pressing enter on the empty list seemingly cancels the build.

**Steps to reproduce**

1. Open a CMake project in Neovim with a CMakePresets.json that has configure presets but no build presets defined.
2. :CMakeBuild
3. Select a configure preset
4. See the empty build preset list
5. Notice how nothing happens and the build does not start.

**Expected behavior**

After prompting for the configure preset, if no build presets are present, the build should just immediately proceed using default options or even the 'build_options' set in the nvim lua config.

Or at the very least, pressing Enter after it asks for build presets when none are available should still let the build proceed.

**Screenshots**

<img width="1045" height="241" alt="Image" src="https://github.com/user-attachments/assets/40c95e5f-1c78-4703-a4a5-a5011e6b6996" />

**Environment**

- OS: Arch Linux
- Desktop environment: Hyprland
- Plugins commit hash: 17244215b1a96e4b2a83a16abd6719197f270f96

PR #334, opened 2025-08-03
Fix: allow build with no selected build preset

bailwillharr:
This pull requests fixes the behaviour I mentioned in Issue #333 (CMake build presets are required when using configure presets). It allows the user to proceed with building if a CMake configure preset has been selected but no build presets are available.

This change was implemented by creating a sentinel object for the `BuildPreset` class which represents the choice of no build preset. This choice **cannot** be specified with `build_preset = nil` as that represents the case where a build preset is yet to be selected.

Code can check for this 'no build preset wanted' state by calling `build_preset.is_none` which returns `true` in that case. 

A few `if` statements in `lua/cmake-tools/init.lua` are modified to ensure that `build_preset` is not used when set to `BuildPreset.None`.

In addition, a minor typo causing undersired behaviour was fixed on line 33 (previously line 21) of `lua/cmake-tools/build_preset.lua`.

### DIFF
--- a/lua/cmake-tools/build_preset.lua
+++ b/lua/cmake-tools/build_preset.lua
@@ -3,6 +3,18 @@ local osys = require("cmake-tools.osys")
 
 local BuildPreset = {}
 
+-- 'None' instance for when no build preset should be used.
+BuildPreset.None = {
+  is_none = true,
+  get_build_target = function()
+    return ""
+  end,
+  get_build_type = function()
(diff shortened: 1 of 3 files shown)


bailwillharr:
Thanks a lot for this fix man. Tbh my experience with Lua is pretty limited so I'm glad someone who knows what they're doing stepped up :).

PR #346, opened 2025-09-29
vimux: simplify exit handling by always returning success

DemiDroL:
The previous implementation relied on terminal.handle_exit(), which expected lock and exit_code files to be created. Vimux does not support this workflow, causing errors when trying to handle exit codes.

This change removes the dependency on terminal.handle_exit() and instead always calls on_exit(0) if provided, treating every run as successful.


lceWolf:
Thank you for your contribution

PR #351, opened 2025-11-01
  Add automatic kit scanning functionality

drook207:
## Summary

  This PR adds automatic CMake kit detection and scanning capabilities to
  cmake-tools.nvim, making it easier to get started without manually
  configuring compiler kits.

  - Automatically scans PATH for GCC and Clang compiler pairs
  - Detects compiler versions and finds matching C++ compilers
  - Discovers linkers (lld/ld) and toolchain files from environment
  - Saves discovered kits to `~/.config/cmake-tools/cmake-kits.json`
  - Falls back to automatic scanning when no kits file exists
  - Adds `CMakeScanForKits` user command for manual scanning

  ## Changes

  - **New module**: `lua/cmake-tools/scanner.lua` - Core kit scanning
  implementation with:
    - Compiler detection (GCC/Clang with version parsing)
    - Linker and toolchain file discovery
    - Custom JSON encoder for saving kits
    - PATH scanning and file utilities

  - **Enhanced workflow**: Automatic kit scanning when `cmake-kits.json`
  doesn't exist

  - **User command**: `CMakeScanForKits` for manual kit discovery

  - **Default config**: Set `cmake_kits_path` to `~/.config/cmake-tools/` by
  default

  ## Test plan

  - [x] Test automatic kit scanning on first run (no kits file)
  - [x] Test manual `CMakeScanForKits` command
  - [x] Verify GCC detection and version parsing
  - [x] Verify Clang detection and version parsing
  - [x] Check that generated kits file is valid JSON
  - [x] Ensure linker and toolchain detection works correctly
  - [x] Test with multiple compilers in PATH


lua/cmake-tools/scanner.lua
```diff
@@ -0,0 +1,211 @@
...
+  return paths
+end
+
+function scanner.get_gcc_version(gcc_path)
+  local success, exit_code, output = scanner.execute_command('"' .. gcc_path .. '" --version')
+  if output == nil then
+    return nil
+  end
+  -- Try multiple patterns to match different gcc output formats
+  local version = output:match("gcc%s+%(GCC%)%s+([%d%.]+)") -- "gcc (GCC) 15.2.1"
```
lceWolf: I think you could match against something like `gcc[%s%a]([%d%.]+)`

lua/cmake-tools/scanner.lua
```diff
@@ -0,0 +1,211 @@
...
+  local base_name = c_compiler:match("([^/\\]+)$")
+  local cxx_name
+  if base_name:match("gcc") then
+    cxx_name = base_name:gsub("gcc", "g++")
+  elseif base_name:match("clang") then
+    cxx_name = base_name:gsub("clang", "clang++")
+  else
+    return nil
+  end
+  local cxx_path = dir .. "/" .. cxx_name
```
lceWolf: In general, it looks like you are expecting a linux system. You should support windows as well. Paths and commands can be different

lua/cmake-tools/scanner.lua
```diff
@@ -0,0 +1,211 @@
+local scanner = {}
```
lceWolf: why isn't this part of cmake-tools/kits.lua?

lua/cmake-tools/scanner.lua
```diff
@@ -0,0 +1,211 @@
...
+    return nil
+  end
+  local cxx_path = dir .. "/" .. cxx_name
+  if scanner.file_exists(cxx_path) then
+    return cxx_path
+  end
+  return nil
+end
+
+function scanner.find_linker_pair(dir, linker_name)
```
lceWolf: this function's name suggests two return values, yet it only returns one

lua/cmake-tools/scanner.lua
```diff
@@ -0,0 +1,211 @@
...
+function scanner.get_clang_version(clang_path)
+  local success, exit_code, output = scanner.execute_command('"' .. clang_path .. '" --version ')
+  if output == nil then
+    return nil
+  end
+  local version_line = output:match("clang version ([%d%.]+)")
+  return version_line
+end
+
+function scanner.find_compiler_pair(dir, c_compiler)
```
lceWolf: this function's name suggests two return values, yet it only returns one

lua/cmake-tools/scanner.lua
```diff
@@ -0,0 +1,211 @@
...
+          toolchainFile = (toolchainFile or ""),
+        }
+        table.insert(kits, kit)
+      end
+    end
+  end
+
+  if #kits == 0 then
+    print("No compilers found in PATH.")
+    return {}
```
lceWolf: should an empty file be created in this case? Otherwise the automatic scan would scan the system everytime

lceWolf (commented):
I like the idea, it replicates vscode's automatic kit detection 👍

lua/cmake-tools/scanner.lua
drook207: I initially started within the kits module, but then I thought it would gain more maintainability and structure to add a separate module. Also in terms of expandability.

plugin/cmake-tools.lua
```diff
@@ -186,6 +186,15 @@ vim.api.nvim_create_user_command(
     desc = "CMake select cmake kit",
   }
 )
+--- CMake scan for kits
+vim.api.nvim_create_user_command(
+  "CMakeScanForKits", -- name
```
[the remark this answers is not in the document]
drook207: Hey together, sorry that my pr is stale at the moment. I'm still working on it but I barely find the time to put my focus on it.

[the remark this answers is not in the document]
drook207: Hey together, sorry that my pr is stale at the moment. I'm still working on it but I barely find the time to put my focus on it.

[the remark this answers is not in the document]
lceWolf: No worries! Take your time. I am glad you are taking care of this topic

drook207:
Hi guys, sorry that it took so long to continue, but as I got back to it i decided to start right from the beginning. I now use the nvim api for all the detection functions and I try to make it as extensible as possible. For now we are able to detect clang and gcc compiler on linux systems (And cross compilation compilers as well). I would suggest to do another PR for the WIndows MSVC and MinGW functionality.

drook207:
And one further question. Should we regenerate the cmake kits file entirely when the user command is triggered or should we append only new kits to the file?

lua/cmake-tools/init.lua
```diff
@@ -214,7 +218,7 @@ function cmake.generate(opt, callback)
 
   -- if exists cmake-kits.json, kit is used to set
   -- environmental variables and args.
-  local kits_config = kits.parse(const.cmake_kits_path, config.cwd)
+  local kits_config = kits.parse(const.cmake_config_path, config.cwd)
```
lceWolf, line 221: I think this has to stay `const.cmake_kits_path`, right?

If so, do we need `const.cmake_config_path` at all?

lua/cmake-tools/const.lua
```diff
@@ -24,7 +24,8 @@ local const = {
     ---@type string|fun(): string
     target = vim.loop.cwd, -- path or function returning path to directory, this is used only if action == "soft_link" or action == "copy"
   },
-  cmake_kits_path = nil, -- this is used to specify global cmake kits path, see CMakeKits for detailed usage
+  cmake_config_path = vim.fn.expand("~") .. "/.config/cmake-tools/", -- this is used to specify global cmake config path
+  cmake_kits_path = vim.fn.expand("~") .. "/.config/cmake-tools/cmake-kits.json", -- this is used to specify global cmake kits path, see CMakeKits for detailed usage
```
lceWolf: could use `stdpath("config")` right?

lua/cmake-tools/init.lua
```diff
@@ -65,6 +66,9 @@ function cmake.setup(values)
   cmake.register_autocmd()
   cmake.register_autocmd_provided_by_users()
   cmake.register_scratch_buffer(config.executor.name, config.runner.name)
+  if not vim.uv.fs_stat(const.cmake_kits_path) then
```
lceWolf, line 69: maye we should add an option like we did for the presets to enable kit usage? 

I think it might be worth keeping `const.cmake_kits_path = nil` as default and not change existing behaviour 🤔

lua/cmake-tools/init.lua
```diff
@@ -65,6 +66,9 @@ function cmake.setup(values)
...
+    scanner.scan_for_kits()
```
lceWolf, line 70: this is a synchronous which can block the ui on startup. I think it is work running async, right?

lua/cmake-tools/scanner.lua
```diff
@@ -0,0 +1,186 @@
...
+  return {
+    c = (prefix or "") .. c_name,
+    cxx = (prefix or "") .. companions.cxx,
+    linker = (prefix or "") .. companions.linker,
+    prefix = prefix or "",
+  }
+end
+
+local function get_path_executables()
+  local path_dirs = vim.split(vim.env.PATH or "", ":", { plain = true })
```
lceWolf: this will break on windows

lua/cmake-tools/types.lua
```diff
@@ -18,6 +18,7 @@ local Types = {
   "ANOTHER_JOB_RUNNING",
   "CMAKE_RUN_FAILED",
   "SETTINGS_ALREADY_OPENED",
+  "CANNOT_FIND_CMAKE_KITS",
```
lceWolf, line 21: this return type is not used, is it?

scripts/scan_kits.lua
```diff
@@ -0,0 +1,11 @@
+-- scripts/scan_kits.lua
```
lceWolf: what is this file for?

codebook.toml
```diff
@@ -0,0 +1 @@
+words = ["cxx"]
```
lceWolf: I guess this file was not meant to be added?

lua/cmake-tools/scanner.lua
```diff
@@ -0,0 +1,186 @@
...
+  if not compiler or compiler == "" then
+    return nil
+  end
+  local version_output = vim.fn.system({ compiler, "--version" })
+  local version = version_output:match("%d+%.%d+%.%d+")
+  return version
+end
+
+-- Main function to scan for kits
+function scanner.scan_for_kits()
```
lceWolf, line 135: This function has quite a few log statements. Are they worth keeping or is it noise for the user?

tests/scanner_spec.lua
```diff
@@ -0,0 +1,106 @@
...
+      end
+    end
+  end)
+end)
+
+describe("kits", function()
+  it("parse from global file", function()
+    local kit = require("cmake-tools.kits")
+    local const = require("cmake-tools.const")
+    local kits = kit.get(const.cmake_kits_path, vim.loop.cwd())
```
lceWolf, line 103: This test requires the other tests to have written the file, making tests order dependent. That should be avoided

tests/scanner_spec.lua
```diff
@@ -0,0 +1,106 @@
+-- tests/scanner_spec.lua
+describe("scanner", function()
```
lceWolf, line 2: These tests have sideeffects which should be avoided:
a) they write to the actual filesystem (config path). 
b) they run `gcc --version` and `clang --version` in a subprocess which introduces a depenceny to the host system and there is not deterministic.

lua/cmake-tools/scanner.lua
```diff
@@ -0,0 +1,186 @@
...
+    if exe == c_name then
+      return "", c_name
+    end
+  end
+  return nil, nil
+end
+local function derive_toolchain(prefix, c_name)
+  local map = {
+    gcc = { cxx = "g++", linker = "ld" },
+    clang = { cxx = "clang++", linker = "lld" },
```
lceWolf, line 58: not sure, but I think clang might also use `ld` in some cases. Have to double check that tho

lua/cmake-tools/scanner.lua
```diff
@@ -0,0 +1,186 @@
+local constants = require("cmake-tools.const")
+local scanner = {}
+
+local C_COMPILERS = { "gcc", "clang" }
+local TOOLCHAIN_SEARCH_PATHS = {
```
lceWolf, line 5: I am not familiar with kits, but AFAIK there is no "default toolchain location" in cmake. If we need to fine the toolchain, we might need to check the CMAKE_TOOLCHAIN_FILE env var, don't we?

lceWolf: Maybe I misunderstand the kits documentation, but I think the toolchain file is part of the kit...
https://vector-of-bool.github.io/docs/vscode-cmake-tools/kits.html#toolchain

lua/cmake-tools/scanner.lua
```diff
@@ -0,0 +1,186 @@
...
+  local version_output = vim.fn.system({ compiler, "--version" })
+  local version = version_output:match("%d+%.%d+%.%d+")
+  return version
+end
+
+-- Main function to scan for kits
+function scanner.scan_for_kits()
+  vim.notify("Scanning for kits…")
+
+  local executables = get_path_executables()
```
lceWolf: I think this could be a module level variable. No need to parse that everytime we `call scan_for_kits`

lua/cmake-tools/scanner.lua
```diff
@@ -0,0 +1,186 @@
...
+      end
+    end
+  end
+  return executables
+end
+local function discover_toolchains(executables)
+  local seen = {}
+  local chains = {}
+
+  for exe in pairs(executables) do
```
lceWolf, line 92: I think this filtering could be moved to the point of searching the paths variable

lua/cmake-tools/scanner.lua
```diff
@@ -0,0 +1,186 @@
...
+      if vim.fn.filereadable(full_path) == 1 then
+        return full_path
+      end
+    end
+  end
+
+  return nil
+end
+local function match_c_compiler(exe)
+  for _, c_name in ipairs(C_COMPILERS) do
```
lceWolf:
```suggestion
  for _, compiler_name in ipairs(C_COMPILERS) do
```
`c_name` is rather hard to understand when not directly associated with `C_COMPILERS`

lua/cmake-tools/scanner.lua
```diff
@@ -0,0 +1,186 @@
...
+  }
+  local companions = map[c_name]
+  if not companions then
+    return nil
+  end
+
+  return {
+    c = (prefix or "") .. c_name,
+    cxx = (prefix or "") .. companions.cxx,
+    linker = (prefix or "") .. companions.linker,
```
lceWolf, line 68: I think this is actually wrong for lld. See [It is always a cross-linker, meaning that it always supports all the above targets however it was built. In fact, we don’t provide a build-time option to enable/disable each target. This should make it easy to use our linker as part of a cross-compile toolchain.](https://lld.llvm.org/)

lua/cmake-tools/scanner.lua
```diff
@@ -0,0 +1,186 @@
...
+        end
+      end
+    end
+  end
+  return chains
+end
+
+local function check_executable_exists(compiler)
+  if not compiler or compiler == "" then
+    return nil
```
lceWolf: based on the function name, this function should return a boolean

lua/cmake-tools/scanner.lua
```diff
@@ -0,0 +1,186 @@
...
+      local kit = { compilers = {} }
+
+      local version = get_compiler_version(tc.c)
+      local prefix_label = tc.prefix ~= "" and (tc.prefix:gsub("%-$", "") .. " ") or ""
+      kit.name = prefix_label .. tc.c .. " " .. (version or "Unknown")
+
+      kit.compilers.C = get_executable_path(tc.c)
+
+      if has_cxx then
+        kit.compilers.CXX = get_executable_path(tc.cxx)
```
lceWolf: I think this can be called unconditionally. `vim.fn.exepath` should yield the correct result

lceWolf:
> Hi guys, sorry that it took so long to continue, but as I got back to it i decided to start right from the beginning. I now use the nvim api for all the detection functions and I try to make it as extensible as possible. For now we are able to detect clang and gcc compiler on linux systems (And cross compilation compilers as well). I would suggest to do another PR for the WIndows MSVC and MinGW functionality.

No worries. Thats quite the topic to tackle and you are moving in a good direction! :) 

> I would suggest to do another PR for the WIndows MSVC and MinGW functionality.

Maybe I missed a detail, but I think the PR might break existing setups on windows since the PATH parsing does only work for Unix systems

lceWolf:
> And one further question. Should we regenerate the cmake kits file entirely when the user command is triggered or should we append only new kits to the file?

How does the VS Code extension handle it?
I think stale kits should be removed

lua/cmake-tools/init.lua
drook207: The parse function expects not only a path but also the filename for the cmake kits. But I agree to leave the `const.cmake_kits_path` nil. 
The questions is should we add the kits filename to the `const` table and concat the config path and the filename ?

lua/cmake-tools/const.lua
drook207: Yes should be fixed with commit e3423fee

scripts/scan_kits.lua
drook207: This should not be commited sorry for the pollution. It was for local tests only.

lua/cmake-tools/scanner.lua
drook207: Yes you are right. I removed it.

lua/cmake-tools/scanner.lua
drook207: Oh okay yes, thanks for the link. Maybe we could still add the ability to check the CMAKE_TOOLCHAIN_FILE env variable on cross compilation kits. For me this would be a big advantage because I am working mostly with embedded linux devices and c++. And I find it quiet annoying to manualy adjust the kits file after detection. 
But yeah that a niche use case.

lua/cmake-tools/scanner.lua
drook207: I don't get what you mean, sorry. Could you maybe give a further explenation?

lua/cmake-tools/scanner.lua
drook207: Ok thanks again for the docs. So should we maybe take out the linker detection complety? Becaus if I get it correct, vscode's cmake-tools does not add the linker at all. It is again a manual step.

drook207:
> > And one further question. Should we regenerate the cmake kits file entirely when the user command is triggered or should we append only new kits to the file?
> 
> How does the VS Code extension handle it? I think stale kits should be removed

No they add and update only. 
[docs](https://vector-of-bool.github.io/docs/vscode-cmake-tools/kits.html#user-local-kits:~:text=Warning%20The%20name,update%20existing%20ones.)

I'm with you, stale kits should be cleaned if one triggers scan for kit manually

drook207:
> > Hi guys, sorry that it took so long to continue, but as I got back to it i decided to start right from the beginning. I now use the nvim api for all the detection functions and I try to make it as extensible as possible. For now we are able to detect clang and gcc compiler on linux systems (And cross compilation compilers as well). I would suggest to do another PR for the WIndows MSVC and MinGW functionality.
> 
> No worries. Thats quite the topic to tackle and you are moving in a good direction! :)
> 
> > I would suggest to do another PR for the WIndows MSVC and MinGW functionality.
> 
> Maybe I missed a detail, but I think the PR might break existing setups on windows since the PATH parsing does only work for Unix systems

Yes you were right, did not see that. I fixed the path separator so the plugin should continue to work on windows. But I will ensure that tomorrow with my work PO

lua/cmake-tools/init.lua
lceWolf: Maybe I misunderstand, but currently the parse function expects a directory and searches that directory for either `cmake-kits.json` or `CMakeKits.json`.

lua/cmake-tools/scanner.lua
lceWolf: Personally I don't use kits, but cmake-presets. There I also have to define the toolchain within the preset by hand. I think an auto-detection is not quite ideal. If detecting kits for the first time, all these kits would then have the same (auto-detected) toolchain-file, which might not be correct. Hence the user would need to fix that anyways. I'd opt for manually setting the toolchain-file as per default (and behave similar to VSCode's plugin)

lua/cmake-tools/scanner.lua
lceWolf: Sounds reasonable 👍

lua/cmake-tools/scanner.lua
lceWolf: Instead of collecting all executables into a set and then filtering for compilers in a second pass, the compiler matching should happen inline during the PATH scan to avoid building a large intermediate table

lua/cmake-tools/scanner.lua
```diff
@@ -0,0 +1,172 @@
...
+
+local executables = get_path_executables()
+
+function scanner.scan_for_kits()
+  local toolchains = discover_toolchains(executables)
+  local kits = {}
+
+  for _, tc in ipairs(toolchains) do
+    local has_c = check_executable_exists(tc.c)
+    local has_cxx = check_executable_exists(tc.cxx)
```
lceWolf, line 141: unused variables (`has_c` too)

PR #354, opened 2025-12-11
fix: initialize buildPresets to empty table before insert

deeedob:
Prevents 'bad argument #1 to insert (table expected, got nil)' crash when CMakePresets.json has no buildPresets key.


deeedob:
Note that it resolves the error but it's still expected to have a `buildPresets` section in your CMakePresets, which isn't required. Without it I can't run a target, as it shows 'None' for selecting a build preset.

PR #355, opened 2025-12-16
Support cmakeSettings field in kits files

mattkrass:
Adding support for the `cmakeSettings` field in CMake Kits files.

### DIFF
--- a/lua/cmake-tools/kits.lua
+++ b/lua/cmake-tools/kits.lua
@@ -133,7 +133,7 @@ function kits.build_env_and_args(kit_name, escape, cwd, global_kits_path)
         end
       else
         if escape then
-          add_args({ "-D" .. k .. ":STRING=" .. '"' .. v '"' })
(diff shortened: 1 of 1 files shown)


lua/cmake-tools/kits.lua
```diff
@@ -123,6 +123,15 @@ function kits.build_env_and_args(kit_name, escape, cwd, global_kits_path)
   if kit.toolchainFile then
     add_args({ "-DCMAKE_TOOLCHAIN_FILE:FILEPATH=" .. kit.toolchainFile })
   end
+  if kit.cmakeSettings then
+    for k, v in pairs(kit.cmakeSettings) do
+      if type(v) == "boolean" then
+        add_args({ "-D" .. k .. ":BOOL=" .. string.upper(string.format("%s", v)) })
+      else
+        add_args({ "-D" .. k .. ":STRING=" .. string.format("%s", v) })
```
lceWolf: I think this needs escaping like the other cases have. 

Also, `string.format` is not needed here, right?

PR #357, opened 2026-01-13
Recursively respect `include` directives in CMake presets files

mous16:
With respect to  #310 , this PR aims to handle `include` directives in CMake presets files, allowing to parse also the directives in included files recursively, and avoiding multiple inclusion of same file.

I've also added some stricter controls on CMakePresets and CMakeUserPresets file names, to avoid wrong deduction of include fields.

I'm not a Lua developer, so the code may not be perfect.  
Please, let me know if it requires some changes to be acceptable.


lceWolf:
Thanks for taking care of this. Much appreciated!

PR #358, opened 2026-01-20
use overseer's default strategy

lceWolf:


lceWolf:
Thanks for testing :)

